Issue #1, opened 2021-02-23
Create package based on CES.jl structure


odunbar:
- [x] We need to create this package with bors, etc (the same as used in CalibrateEmulateSample.jl)
- [x] We need a readme.md for the front page
- [x] We need to populate the Project toml with the packages from CalibrateEmulateSample.jl required (only) for EnsembleKalmanProcesses (This will not include the python dependencies)
- [x] We need to copy across the Loss Minimization and to copy-and-cut the optimization part of Lorenz 96/Cloudy examples.

jakebolewski:
@odunbar code coverage should be enabled now

jakebolewski:
Closed by #9

PR #3, opened 2021-02-24
Initial move of source files and runtests from CalibrateEmulateSample

odunbar:
## Purpose
Resolves some of #1 

## Contains
We move the following modules from CES
- `EnsembleKalmanProcesses`, renamed `EnsembleKalmanProcessModule` (for obvious name clash)
- `Observations`
- `DataStorage`
- `ParameterDistributionStorage`

I also have moved over all unit tests for these modules. and updated the Project toml


odunbar:
Copy pasted from CES, without edit, so I will merge without review

PR #4, opened 2021-02-25
examples from CES, Calibration stage only

odunbar:
## Purpose
Addresses part of #1 
Get examples from the `CalibrateEmulateSample.jl` repository, keeping only the calibration stage - I have added plots as in LossMinimization to the other examples

## Contains
- LossMinimization
- Lorenz 96 - calibration only
- Cloudy - calibration only


odunbar:
Primarily copy-paste from CES so I will merge this initial examples commit without review.

PR #7, opened 2021-03-04
Add ClimateMachine example

ilopezgp:
This PR adds functions needed to run a simulation of ClimateMachine.jl with EKP.jl, as well as bash scripts for parallel runs of ClimateMachine.jl within each ensemble. The bash scripts assume the use of a SLURM workload manager (e.g. Caltech Central HPC).

The simulation chosen in this case is a stable boundary layer simulation using a one-dimensional anelastic Atmos model. This example is not self-contained, in that it requires installation of ClimateMachine.jl, similarly to the Cloudy.jl example.


ilopezgp:
@odunbar The PR is in good shape already, but I do not have a separate Project.toml for the example. Could you walk me through the process of setting one up?

examples/ClimateMachine/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+The example makes use of a simple single column model configuration with two tunable 
+parameters to showcase the use of EnsembleKalmanProcesses.jl with ClimateMachine.jl.
+This example is a perfect model experiment, in the sense that the ground truth is 
+generated using the same model and a certain combination of parameters.
+
+To run the example using a SLURM workload manager, simply do:
+
+>> sbatch ekp_calibration.sbatch
+
```
odunbar, line 18: Maybe state a little more here, such as
```
This will create a queue on the workload manager running 
1. ekp_init_calibration (creating initial parameters)
Followed by iterating between
2a. ekp_single_cm_run (running the GCM)
2b. ekp_cont_calibration (updating parameters with EKP)
until the desired number of iterations (defined in ekp_calibration.sbatch) are complete.
```

odunbar: RE the truth output here - 
1. Is it something that the user can generate themselves from the true parameters? If so, then we don't require the nc data file upload on this repo, and instead we should have a script to generate the truth? 
2. If it is instead something that the users cannot easily generate then maybe just say here in the readme that we used parameters `(c_1,c_2) = (...)` to generate the true data stored in `truth_output/SBL_ANELASTIC_1D_AtmosLESDefault.nc`, it is the goal of EKP to recover these.

odunbar: Otherwise LGTM - if you confirmed the examples run?

odunbar:
Thanks for clarifying the ground truth exps in the readme, LGTM

PR #8, opened 2021-03-08
Unscented Kalman Inversion

Zhengyu-Huang:
Implement the unscented Kalman inversion [1] in the EKP library

The method is a general derivative-free optimization approach for inverse problems

- give both mean and covariance approximation of the posterior distribution with an uninformative prior
- require 2Nθ + 1 particles in general
- no empirical variance inflation or early stopping needed
- no ensemble collapse
- the 3-sigma confidence interval covers the truth parameters for perfect models 

Several examples are presented
1. Cloudy example
    The convergence of the mean and 3-sigma of the three transformed variables and the true values are 

Cloudy

2. Lorenz96 example
     The convergence of the mean and 3-sigma of the three transformed variables and the true values are

Lorenz96

1. Huang, Daniel Z., Tapio Schneider, and Andrew M. Stuart. "Unscented kalman inversion." arXiv preprint arXiv:2102.01580 (2021).


odunbar:
This looks great! Thanks Daniel, let's have a chat about incorporating this into the EKP framework. I see 2 parts to it (Please others chime in if you disagree):

1. We split the  `UnscentedKalmanProcess` struct you have made into (1) what is specific to UKI, e.g UKI weights, and (2) what is general to the problem e.g the observation covariance.  Then put everything from (1) into a struct that looks like
```Julia
struct  Unscented <: Process
    uki_param1::P1
    uki_param2::P2
end
```
Everything in (2) can be placed into the `EnsembleKalmanProcess` struct as seen in `EnsembleKalmanProcesses`.

2. The general EKP functions should be written to be dispatched on the `Process` object e.g the update function should be written as 
```Julia
update_ensemble!(ekp::EnsembleKalmanProcess{FT, IT, Unscented},...)
```

It's fine for you to keep everything in the current file you have.

Zhengyu-Huang:
> This looks great! Thanks Daniel, let's have a chat about incorporating this into the EKP framework. I see 2 parts to it (Please others chime in if you disagree):
> 
> 1. We extract from the  `UnscentedKalmanProcess` all the general parameters (e.g the observation covariance), and  then we can create the `EnsembleKalmanProcess{FT,IT, <:Process}` struct as seen in `EnsembleKalmanProcesses`. We require a new process type, and so the rest of the parameters from `UnscentedKalmanProcess` should be specific to UKI, (e.g UKI weights), and we put them in here:
> 
> ```julia
> struct  Unscented <: Process
>     uki_param1::P1
>     uki_param2::P2
> end
> ```
> 
> 1. The general EKP functions should be written to be dispatched on the `<:Process` objects e.g your new update function should be called as
> 
> ```julia
> update_ensemble!(ekp::EnsembleKalmanProcess{FT, IT, Unscented},...)
> ```
> 
> It's fine for you to keep everything in the current file you have. Let's see how easily we can do this!
> 
> PS. you call it an optimization method - is it not an approximate UQ method?

@odunbar I have updated with Unscented process, could you have a look?

odunbar (changes requested):
It's looking better! We need to update the function so that it is more like 
`update_ensemble!(uki, u^{n}, G(u^{n}))` and the output is `u^{n+1}`. 

I think this is possible by setting ourselves up to calculate 
1. In the constructor, for EKP copy-paste the lines 558-584: and define the `init_params = construct_sigma_ensemble(process, u_p_mean, uu_p_cov)` (as this doesn't rely upon the model right?)
2. remove applying G within `update_ensemble` and instead apply G to u_p in the outer loop (in the example script) and pass in the data matrix "G(u_p)" into update_ensemble
3. update_ensemble now effectively starts at line 589-610, and then move the block 558-584 to the end of this currently written function so that it ends with creating then new `u_p`.

If this doesn't make sense I'm happy to meet. But it looks good!

odunbar:
Well in complex applications we may not be able to simply "call G" in the update ensemble, that's the issue. I agree your current implementation is nice and clean but this code is likely going to be hooked to possibly nasty models (e.g the climate model scenario). 

The way we handle it with the blackbox code is:
1) initialize (and output ensemble parameters) i.e generate a collection of parameter values
2) pass ensemble parameters to black box to generate data
3) Call update_ensemble on the data to produce a new ensemble of parameters, go back to 2)
END) iterate 2-3 until desired convergence/iterations

step 2) could be literally anything, in the GCM setting it often requires writing parameters into run-files, then setting off ensemble of simulations on HPC that reads these different files, then postprocessing all the data. Working with parameters and data, rather than distributions and functions, is just more flexible here.

step 2) takes (\theta^n_i) -> G(\theta^n_i) and 
step 3) takes (G(\theta^n_i) -> \theta^{n+1}_i

Zhengyu-Huang:
> Well in complex applications we may not be able to simply "call G" in the update ensemble, that's the issue. I agree your current implementation is nice and clean but this code is likely going to be hooked to possibly nasty models (e.g the climate model scenario).
> 
> The way we handle it with the blackbox code is:
> 
> 1. initialize (and output) ensemble parameters i.e generate a collection of parameter values
> 2. pass ensemble parameters to black box to generate data
> 3. Call update_ensemble on the data to produce a new ensemble of parameters, go back to 2)
> 
> END) iterate 2-3 until desired convergence/iterations
> 
> step 2) could be literally anything, in the GCM setting it often requires writing parameters into run-files, then setting off ensemble of simulations on HPC that reads these different files (currently based on bash and Slurm scripts), then postprocessing all the data. Working with parameters and data, rather than distributions and functions, is just more flexible here.
> 
> step 2) takes (\theta^n_i) -> G(\theta^n_i) and
> step 3) takes (G(\theta^n_i) -> \theta^{n+1}_i

Ok, I updated it , could you have a look?

examples/Cloudy/Cloudy_example_ukp.jl
```diff
@@ -0,0 +1,278 @@
...
+dummy = 1.0
+dist_type = PDistributions.Gamma(dummy, dummy, dummy)
+g_settings = GModel.GSettings(kernel, dist_type, moments, tspan)
+
+
+
+err = zeros(N_iter) 
+for i in 1:N_iter
+    # prediction step 
+    u_p = EnsembleKalmanProcessModule.update_ensemble_prediction!(ukiobj)
```
odunbar, line 196: `u_p` can be calculated from within the ensemble update step. If you still want to retain the conceptual setup of `analysis` step and `prediction` step. Then I would suggest the following.
Define the function `update_ensemble!` as
```julia
function update_ensemble!(ukiobj::EnsembleKamanProcess{..,Unscented}, g_ens}
    #perform analysis on the model runs
    update_ensemble_analysis!(ukiobj,g_ens)
    #perform new prediction output to model parameters u_p
    return update_ensemble_prediction!(ukiobj.process) 
end
```
Note we don't want `update_ensemble_prediction` to depend on `ukiobj` as we want to use this function in the constructor of the ukiobj. (It can however depend on the process)

src/EnsembleKalmanProcess.jl
```diff
@@ -336,4 +337,324 @@ function update_ensemble!(ekp::EnsembleKalmanProcess{FT, IT, Sampler{FT}}, g_in:
...
+
+# outer constructors
+function EnsembleKalmanProcess(
+    obs_mean::Array{FT, 1},
+    obs_noise_cov::Array{FT, 2},
+    process::Unscented{FT, IT};
+    Δt=FT(1)) where {FT<:AbstractFloat, IT<:Int}
+
+#initial parameters stored as columns
+init_params=[]
```
odunbar, line 375: Here we want to create the initial sigma ensemble from the process. we can define in this function
```Julia
init_params = update_ensemble_prediction!(process)
```

src/EnsembleKalmanProcess.jl
```diff
@@ -336,4 +337,324 @@ function update_ensemble!(ekp::EnsembleKalmanProcess{FT, IT, Sampler{FT}}, g_in:
...
+        xy_cov .+= cov_weights[i]*(x[:, i] - x_mean)*(obs_mean[:, i] - y_mean)'
+    end
+    
+    return xy_cov
+end
+
+"""
+uki prediction step : generate sigma points
+"""
+function update_ensemble_prediction!(uki::EnsembleKalmanProcess{FT, IT,Unscented}) where {FT<:AbstractFloat, IT<:Int}
```
odunbar, line 551: As noted before we don't actually need the whole object `ukiobj` here, we only need `process` as far as I can see? So we can rewrite as 
```Julia
function update_ensemble_prediction!(process::Unscented) 
```

src/EnsembleKalmanProcess.jl
```diff
@@ -336,4 +337,324 @@ function update_ensemble!(ekp::EnsembleKalmanProcess{FT, IT, Sampler{FT}}, g_in:
...
+    end
+    
+    u_mean  = uki.process.u_mean[end]
+    uu_cov = uki.process.uu_cov[end]
+    
+    α_reg = uki.process.α_reg
+    r = uki.process.r
+    Σ_ω = uki.process.Σ_ω
+    
+    N_u, N_ens = length(uki.process.u_mean[1]),  uki.N_ens
```
odunbar, line 566: By the looks of it, we don't even use `N_ens` in this function. so this can be removed

odunbar (changes requested):
This should be the last iteration and then I think we are done. Although we will need some runtests for this.

src/EnsembleKalmanProcess.jl
Zhengyu-Huang: updated! Thanks!

src/EnsembleKalmanProcess.jl
Zhengyu-Huang: updated thank!

src/EnsembleKalmanProcess.jl
Zhengyu-Huang: updated. Thanks

src/EnsembleKalmanProcess.jl
Zhengyu-Huang: updated, thanks!

examples/Cloudy/Cloudy_example_ukp.jl
Zhengyu-Huang: Updated!

src/EnsembleKalmanProcess.jl
```diff
@@ -336,4 +337,336 @@ function update_ensemble!(ekp::EnsembleKalmanProcess{FT, IT, Sampler{FT}}, g_in:
...
+    c_weights = zeros(FT, N_u)
+    mean_weights = zeros(FT, N_ens)
+    cov_weights = zeros(FT, N_ens)
+    
+    # todo parameters λ, α, β
+    
+    κ = 0.0
+    β = 2.0
+    α = min(sqrt(4/(N_u + κ)), 1.0)
+    λ = α^2*(N_u + κ) - N_u
```
odunbar, line 433: The parameters `kappa` `beta` are essentially magic numbers. they should either be replaced with actually `0.0` and `2.0` or they should be parameters - perhaps with these default values (that can be changed by a user, and have some brief explanation).

odunbar (commented):
Thanks for the latest revision!
The source code / examples look good to me. (i have one small comment)

All we need now are the `runtests`, So we need something to run through your source functions so the codecov is happy. Let me know if you want to chat about how to set these up.

src/EnsembleKalmanProcess.jl
```diff
@@ -336,4 +337,336 @@ function update_ensemble!(ekp::EnsembleKalmanProcess{FT, IT, Sampler{FT}}, g_in:
...
+N_ens = 2N_u + 1 #stored with data as columns
+#store for model evaluations
+g=[] 
+# error store
+err = FT[]
+# timestep store
+Δt = Array([Δt])
+
+EnsembleKalmanProcess{FT, IT, Unscented}(init_params, obs_mean, obs_noise_cov, N_ens, g,
+          err, Δt, process)
```
bielim, line 388: Missing indentation

bielim:
A minor thing: Could the ordering be changed such that the `Unscented` process and its outer constructor are moved to the other `Processes` (Inversion, Sampler) at the top of the module? Likewise, the `EnsembleKalmanProcess` and all its constructors should be grouped together in one place.

bielim:
In the construction of an `EnsembleKalmanProcess` of process type `Unscented`, both the initialization of the `Unscented` object and the initialization of the `EnsembleKalmanProcess` (using that `Unscented` object) take the mean and covariance of the observational noise as inputs, which seems a bit redundant. Also, as far as I can see, the `\Delta t` argument to `EnsembleKalmanProcess` remains unused and can thus be confusing. I get that the idea was for `EnsembleKalmanProcess` to take some general arguments such as observation covariance as well as the process type, which is a struct holding all the information that is specific to the process (`Inversion`, `Sampler`, or `Unscented`). However, I wonder if it might be easier to get rid of the "general arguments" (since they might lead to redundancy or confusion) and instead include all of them in the process type structs, such that the construction of a new process would be done simply by `ek_process = EnsembleKalmanProcess(ek_type_struct::Union{Inversion, Sampler, Unscented})`.

bielim:
Also: I think that we should have a separate "cosmetic" PR that takes care of things like proper doc strings for functions (unless it's totally obvious what they do, e.g., getters and setters), making variable names consistent (e.g., the variables denoting the dimension of the data include "N_y", "data_dim", and "output_dim"), etc., in the entire module. Establishing these conventions will make things easier down the road, as people typically use the existing code base as a starting point/template for developing their contributions.

src/EnsembleKalmanProcess.jl
Zhengyu-Huang: Updated

src/EnsembleKalmanProcess.jl
Zhengyu-Huang: updated

Zhengyu-Huang:
> Also: I think that we should have a separate "cosmetic" PR that takes care of things like proper doc strings for functions (unless it's totally obvious what they do, e.g., getters and setters), making variable names consistent (e.g., the variables denoting the dimension of the data include "N_y", "data_dim", and "output_dim"), etc., in the entire module. Establishing these conventions will make things easier down the road, as people typically use the existing code base as a starting point/template for developing their contributions.

Yes, we should have a docstrings PR in to make these more clear for users

Zhengyu-Huang:
> A minor thing: Could the ordering be changed such that the `Unscented` process and its outer constructor are moved to the other `Processes` (Inversion, Sampler) at the top of the module? Likewise, the `EnsembleKalmanProcess` and all its constructors should be grouped together in one place.

Updated, thanks!

Zhengyu-Huang:
> In the construction of an `EnsembleKalmanProcess` of process type `Unscented`, both the initialization of the `Unscented` object and the initialization of the `EnsembleKalmanProcess` (using that `Unscented` object) take the mean and covariance of the observational noise as inputs, which seems a bit redundant. Also, as far as I can see, the `\Delta t` argument to `EnsembleKalmanProcess` remains unused and can thus be confusing. I get that the idea was for `EnsembleKalmanProcess` to take some general arguments such as observation covariance, as well as the process type, which is a struct holding all the information that is specific to the process (`Inversion`, `Sampler`, or `Unscented`). However, I wonder if it might be easier to get rid of the "general arguments" (since they might lead to redundancy or confusion) and instead include all of them in the process type structs, such that the construction of a new process would be done simply by `ek_process = EnsembleKalmanProcess(ek_type_struct::Union{Inversion, Sampler, Unscented})`.

I am not sure how to modify this one, I just tried to make unscented Kalman inversion constructor look like the others, and reuse these data structures

Zhengyu-Huang:
> Thanks for the latest revision!
> The source code / examples look good to me. (i have one small comment)
> 
> All we need now are the `runtests`, So we need something to run through your source functions so the codecov is happy. Let me know if you want to chat about how to set these up.

Updated a calibration test (3) in the runtest

odunbar:
> Also: I think that we should have a separate "cosmetic" PR that takes care of things like proper doc strings for functions (unless it's totally obvious what they do, e.g., getters and setters), making variable names consistent (e.g., the variables denoting the dimension of the data include "N_y", "data_dim", and "output_dim"), etc., in the entire module. Establishing these conventions will make things easier down the road, as people typically use the existing code base as a starting point/template for developing their contributions.

@bielim if you could create issue(s) for this? Certainly I am hoping we can do a docs push in the next couple of for EKP and CES which will i hope will help with both writing docstrings and enforcing more consistency as we get more eyes looking over it

bielim:
> > Also: I think that we should have a separate "cosmetic" PR that takes care of things like proper doc strings for functions (unless it's totally obvious what they do, e.g., getters and setters), making variable names consistent (e.g., the variables denoting the dimension of the data include "N_y", "data_dim", and "output_dim"), etc., in the entire module. Establishing these conventions will make things easier down the road, as people typically use the existing code base as a starting point/template for developing their contributions.
> 
> @bielim if you could create issue(s) for this? Certainly I am hoping we can do a docs push in the next couple of for EKP and CES which will i hope will help with both writing docstrings and enforcing more consistency as we get more eyes looking over it

Done!

Zhengyu-Huang:
> > Also: I think that we should have a separate "cosmetic" PR that takes care of things like proper doc strings for functions (unless it's totally obvious what they do, e.g., getters and setters), making variable names consistent (e.g., the variables denoting the dimension of the data include "N_y", "data_dim", and "output_dim"), etc., in the entire module. Establishing these conventions will make things easier down the road, as people typically use the existing code base as a starting point/template for developing their contributions.
> 
> @bielim if you could create issue(s) for this? Certainly I am hoping we can do a docs push in the next couple of for EKP and CES which will i hope will help with both writing docstrings and enforcing more consistency as we get more eyes looking over it

Any further change requests? Should I rebase and merge?

src/EnsembleKalmanProcess.jl
```diff
@@ -336,4 +368,317 @@ function update_ensemble!(ekp::EnsembleKalmanProcess{FT, IT, Sampler{FT}}, g_in:
...
+    u_mean = Array{FT,1}[]  # array of Array{FT, 2}'s
+    push!(u_mean, u0_mean) # insert parameters at end of array (in this case just 1st entry)
+    uu_cov = Array{FT,2}[] # array of Array{FT, 2}'s
+    push!(uu_cov, uu0_cov) # insert parameters at end of array (in this case just 1st entry)
+    
+    obs_pred = Array{FT, 1}[]  # array of Array{FT, 2}'s
+    err = FT[]
+    
+    Σ_ω = (2 - α_reg^2)*uu0_cov
+    Σ_ν = 2*obs_noise_cov
```
odunbar, line 465: Relating to Melanie's comment of requiring both the observational noise in `EKP` and `Unscented` constructors. It turns out we really don't need these in the `Unscented` object. Note in the constructor it is only used: 
1. `[428] N_y = size(obs_mean, 1)`
2. `[465] Σ_ν = 2*obs_noise_cov`
In the case of 1. we can just provide `N_y` directly. In the case of 2. We could just store the constant "2" and multiply later by the observational noise covariance (stored in `EKP`)- as this matrix is only used in the analysis step of the update.

src/EnsembleKalmanProcess.jl
```diff
@@ -336,4 +368,317 @@ function update_ensemble!(ekp::EnsembleKalmanProcess{FT, IT, Sampler{FT}}, g_in:
...
+
+
+"""
+uki analysis step 
+g is the predicted observations  Ny  by N_ens matrix
+"""
+function update_ensemble_analysis!(uki::EnsembleKalmanProcess{FT, IT,Unscented}, u_p::Array{FT, 2}, g::Array{FT, 2}) where {FT<:AbstractFloat, IT<:Int}
+    
+    obs_mean = uki.obs_mean
+    Σ_ν = uki.process.Σ_ν
```
odunbar, line 591: If we store `Σ_ν` as the constant `2` in the `Unscented` object instead of `2*obs_cov_noise`, then this can simply be replaced by 
`Σ_ν = uki.process.Σ_ν * uki.obs_noise_cov`

odunbar:
LGTM! Thanks

odunbar (approved):

bielim (approved):

PR #27, opened 2021-05-14
Fix EKI update, ensure consistency for stochastic models.

ilopezgp:
This PR fixes the scale of the error used in the EKI update, and adds a flag for deterministic vs stochastic model updates. 

The names of dimensionality of vectors are also updated to be consistent with the glossary.


odunbar:
LGTM!

odunbar (approved):

PR #28, opened 2021-05-18
[WIP] Add documentation for Cloudy example

bielim:
This PR adds documentation for the Cloudy example and fills a part of the documentation needs identified in #15.

- [x] Make Cloudy use the current master of Cloudy rather than the super-old V0.1.0
- [x] Make `Cloudy_example_eki.jl` consistent with terminology listed in the glossary
- [x] Make `Cloudy_example_ukp.jl` consistent with terminology listed in the glossary
- [x] Add documentation for Cloudy example
- [x] Review and finalize docs


docs/src/examples/Cloudy_example.md
```diff
@@ -0,0 +1,82 @@
+# Cloudy Example
+
+### What Is This Example About?
+
+This example is based on Cloudy, a microphysics model that simulates how cloud droplets collide and coalesce into larger drops. Collision-coalescence is a crucial process for the formation of rain. 
+
+Cloudy takes two inputs: The initial mass distribution of the cloud droplets, and a so-called collision-coalescence kernel, which specifies the probability that two droplets of masses x and y collide and coalesce in a given time interval.
```
ilopezgp, line 7: Can you write x and y in math mode?

docs/src/examples/Cloudy_example.md
```diff
@@ -0,0 +1,82 @@
...
+
+### What Is This Example About?
+
+This example is based on Cloudy, a microphysics model that simulates how cloud droplets collide and coalesce into larger drops. Collision-coalescence is a crucial process for the formation of rain. 
+
+Cloudy takes two inputs: The initial mass distribution of the cloud droplets, and a so-called collision-coalescence kernel, which specifies the probability that two droplets of masses x and y collide and coalesce in a given time interval. 
+Cloudy simulates the evolution of the initial cloud droplet mass distribution over time, as more and more droplets combine into bigger drops. It does this not by simulating each individual droplet, but by using a "bulk approach", which only keeps track of some number of moments of the cloud droplet mass distribution.
+
+This example shows how Ensemble Kalman methods can be used to learn the parameter of the initial cloud droplet mass distribution from observations of the moments of that mass distribution at a later time. The collision-coalescence kernel is assumed to be known, but one could also learn the parameters of the kernel instead of the parameters of the droplet distribution (or both).
```
ilopezgp, line 10: _can be used to learn the parameter_ parameter or parameters?

docs/src/examples/Cloudy_example.md
```diff
@@ -0,0 +1,82 @@
...
+lets you install the current master branch):
+```
+pkg > add Cloudy#master
+```
+
+### Structure
+
+The file `DynamicalModel.jl` provides the functionality to run Cloudy, i.e., to map the parameters of the initial cloud droplet mass distribution to the (zeroth-, first-, and second-order) moments of the distribution at a specified end time. In this example, there is no post-processing required to produce the observations from the model output (i.e., the observation map ``\mathcal{H}`` mentioned in the [docs](https://clima.github.io/EnsembleKalmanProcesses.jl/dev/ensemble_kalman_inversion/) is the identity map).
+
+The file `Cloudy_example_eki.jl` sets up the inverse problem and solves it using [Ensemble Kalman inversion](https://clima.github.io/EnsembleKalmanProcesses.jl/dev/ensemble_kalman_inversion/), and the file `Cloudy_example_uki.jl` does the same using unscented Kalman inversion.
```
ilopezgp, line 29: Can you either capitalize Ensemble and Unscented or neither?

docs/src/examples/Cloudy_example.md
bielim: Done!

docs/src/examples/Cloudy_example.md
bielim: Thanks for catching, I went for "ensemble".

docs/src/examples/Cloudy_example.md
bielim: Following the glossary it should be "parameters", you're right :)

ilopezgp (approved):

odunbar (commented):
My summary is maybe that the focus of this is on Cloudy, and running the example. The inverse problem needs less detail here maybe - just define the prior, data and noise clearly! Just because we do have a methods section and other docs for these!

Comments as I see them:
1. Paragraph 2: suggestion: "Cloudy is initialized with a mass distribution of the cloud droplets; this distribution is then evolved in time by using a collision-coalescence kernel to represents the 2-particle interactions in a given time interval. The evolution is completed described by the shape of the initial distribution and the form of the kernel."
2. Write out the formula here in the top section for the distribution and the kernel in this example
3. How many moments are tracked - specify here. It is also not clear to me (yet) if cloudy is tracking only the moments, or if it evolves the distribution (of which you select N moments as your "data")
4. "in realistic applications..." maybe you want to say: "... this parameter estimation procedure using Cloudy will be able to use a measurement of current droplet moments to obtain an the droplet mass distribution at a previous time".
4. Prerequisites: Does cloudy have a release btw? Do we not have prereqs installed already on the EKP github?
5. Structure: 
5. Inverse Problem: Is the Gamma distribution completely described in 3 moments?
6. If we discuss the G map at the start, and the form of the inverse problem in methods sections of the docs , maybe focus here on priors for the parameters, Cloudy output, and the noise (e.g did you want to say what values of time T we use, and what values in the example do we take for the priors, and noise) also you can use some the code names too?
7. We discuss the constrained-unconstrained space in detail in the method section and prior section - so can be more concise here. or can link other pages
8.  I like the solution/output/playing around sections. nice categories for users

bielim:
> My summary is maybe that the focus of this should be on Cloudy, and running the example. The inverse problem needs less detail here maybe - just define the prior, data and noise nice and clearly. Just because we do have a methods section and other docs for these!
> 
> Comments as I see them:
> 
> 1. Paragraph 2: suggestion: "Cloudy is initialized with a mass distribution of the cloud droplets; this distribution is then evolved in time by using a collision-coalescence kernel to represents the 2-particle interactions in a given time interval. The evolution is completed described by the shape of the initial distribution and the form of the kernel."

I changed the original phrasing to:
_Cloudy is initialized with a mass distribution of the cloud droplets; this distribution is then evolved in time, with more and more droplets colliding and combining into bigger drops according to the droplet-droplet interactions specified by a collision-coalescence kernel. The evolution is completely determined by the shape of the initial distribution and the form of the kernel._

> 2. Write out the formula here in the top section for the distribution and the kernel in this example

I changed the structure of the document, such that there is now an "Overview" section followed by a "What Does Cloudy Do?" section. The formulas for the distribution and the kernel are written out in the latter section.

> 3. How many moments are tracked - specify here. It is also not clear to me (yet) if cloudy is tracking only the moments, or if it evolves the distribution (of which you select N moments as your "data")

I tried to make this clearer by describing in some more detail how Cloudy works (section "What Does Cloudy Do?"): Cloudy only tracks three moments of the distribution which -- under the assumption that the distribution retains its Gamma shape -- is sufficient to determine the distribution parameters.

> 4. "in realistic applications..." maybe you want to say: "... this parameter estimation procedure using Cloudy will be able to use a measurement of current droplet moments to obtain an estimated droplet mass distribution at a previous time".

I changed this to: 
_In more realistic applications, this parameter estimation procedure will use actual measurements of cloud properties to obtain an estimated droplet mass distribution at a previous time._

> 5. Prerequisites: Does cloudy have a release btw? Do we not have prereqs installed already on the EKP github (project.toml)?

Yeah I don't know how to deal with this. It is true that the prerequisites are listed in `Project.toml`, but for this example I would like the user to install the current master branch of Cloudy (I don't know how to specify that in the .toml file). Cloudy does have a release (V0.1.0), but it is really old (that's why I wanted to use the master branch).

> 6. Inverse Problem: Is the Gamma distribution completely described in 3 moments?

Yes, the Gamma moments and parameters are related such that you can uniquely determine the parameters from the first three moments of the distribution (for "regular" Gamma distributions you only need the first and second moment, but here we need three because we're working with scaled probability distributions, where the zeroth moment is not 1 but corresponds to the number of cloud droplets.

> 7. If we discuss the G map at the start, and the form of the inverse problem in methods sections of the docs , maybe focus here on priors for the parameters, Cloudy output, and the noise (e.g did you want to say what values of time T we use, and what values in the example do we take for the priors, and noise) also you can use some the code names too?

The choice of priors and covariance as well as the Cloudy output are now described in more detail.

> 8. We discuss the constrained-unconstrained space in detail in the method section and prior section - so can be more concise here. or can link other pages

I included a link, but didn't delete the part about constrained-unconstrained altogether, since we need it when talking about priors (and sometimes a bit of repetition doesn't hurt :))

> 9. I like the solution/output/playing around sections. nice categories for users
> 
> PS Thanks a load for being the guinea pig with regards to how to structure and put information into these sections! You are helping everyone!

My pleasure :-D

odunbar:
Looks great - very clear, lots of information for users, focused on the example.

What do you think about moving the Prerequisites and Structure to the very top of the page? You would need to move the sentence about the dynamical model into the end of "What does cloudy do?" section.

Otherwise, LGTM! Thanks!

bielim:
> Looks great - very clear, lots of information for users, focused on the example.
> 
> What do you think about moving the Prerequisites and Structure to the very top of the page? You would need to move the sentence about the dynamical model into the end of "What does cloudy do?" section.
> 
> Otherwise, LGTM! Merge either way

Yes, I was actually going back and forth between introducing Cloudy first vs. focusing on the code first, so the fact that you suggest putting the "code part" first definitely tips the scale in favor of this :-)  
I've moved the sections on Prerequisites, Structure, as well as the section on Running the Example to the top of the document (right after the Overview), so that the information on how to get started is front and center, before we get into the more theoretical parts.

PR #29, opened 2021-05-22
Fix get_mean() for PDs, fix installation docs

ilopezgp:
The current output of get_mean(ParameterDistribution) is an `Array{FT, 2}` that cannot be used as an input to the `Sampler()`. I did not see any instance of the code where this structure was useful, so I have changed the function to return an `Array{FT, 1}`. 

Note that this change simplifies every single use of get_mean(ParameterDistribution) in the tests and examples.

Apart from this, I have corrected the installation guide.

### DIFF
--- a/docs/src/installation_instructions.md
+++ b/docs/src/installation_instructions.md
@@ -3,8 +3,8 @@
 To build the top-level project, first clone the repository, then instantiate:
 
 ```
-cd CalibrateEmulateSample.jl
-julia --project -e 'using Pkg; Pkg.instantiate()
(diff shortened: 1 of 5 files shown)


odunbar:
We store samples as oriented 2D arrays, viewed as lists of columns e.g if you have samples of data they are always stored as columns, I expect this was the mindset as to why the mean should be a column. However as the statistics `mean` function naturally reduces the dimension, it is definitely sensible to store it as a `Vector{FT}` and I believe we do also this for other "means" in the code too.

so LGTM

odunbar (approved):

PR #30, opened 2021-05-23
Simplify EKI update, correct dims error in nondeterministic settings

ilopezgp:
This PR simplifies substantially the update of the EKI algorithm.

- The computation of the cross-covariance matrices was done through a loop, which is unnecessary since we are already loading the Statistics module.
- Everything was transposed for some reason. Now all data are stored in columns, similarly to the rest of the code.

 The new construction produces the same results (tested with an `assert` before this PR). Note that I have not added the Bessel correction to the cross-covariances `cov_ug` and `cov_gg` to ensure consistency with the papers and the previous version. We can discuss this.

In addition, I have fixed an issue with the construction of `y`, which was `Array{FT,1}` for `!deterministic_forward_map`.

### DIFF
--- a/src/EnsembleKalmanProcess.jl
+++ b/src/EnsembleKalmanProcess.jl
@@ -257,12 +257,12 @@ function find_ekp_stepsize(ekp::EnsembleKalmanProcess{FT, IT, Inversion}, g::Arr
 end
 
 """
-    update_ensemble!(ekp::EnsembleKalmanProcess{FT, IT, <:Process}, g_in::Array{FT,2} cov_threshold::FT=0.01, Δt_new=nothing) where {FT, IT}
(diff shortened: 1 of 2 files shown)


odunbar:
LGTM! Thanks.

As for responses:
1. i didn't realize that the cov(u,g,...) existed when first implementing this (i almost wouldn't be surprised if it didn't!)
2. The transpose was there because initially there was a clash in different parts of the code then we unified everything, and the simplest fix at the time was to just permute things going into this function. 
3. With the `y` construction, i guess it wasn't an issue before because of broadcasting? either way this is fine with me

odunbar (approved):

ilopezgp:
Yes, I think the `cov` functionality for cross-covariances might be a new implementation. And I agree that this was difficult to do when unifying all notation everywhere, which already required a lot of work.

PR #31, opened 2021-05-28
Add documentation for Lorenz 96 example

mhowlan3:


odunbar:
Hi Mike, 

Can you add the new `Make.jl` file so the docs compile online. Also you seem to have a few extra `jld2` files

odunbar:
Sorry it took me so long: Comments as I see them:
1. Overview - describe what the example does. E.g here is L96 with additional periodic-in-time forcing, we try to determine parameters (Amplitude and stationary component) from some [define which] statistics.
2.L96 - Characteristic time = Lyapunov time?
3. L96 -  mention that we stay within the chaotic regime for F?
4. Dynamic - could you define the defaults as you do in the next sections
5. Priors, "wide" is defined with respect to the L96 variability? It may not be wide if one prescribes a high obs noise

Otherwise this looks great! Thanks Mike - it is very clear and nicely partitioned!

PR #40, opened 2021-08-20
Update installation instructions

navidcy:
- Add tagged version installation
- Fix instructions for building documentation

### DIFF
--- a/docs/src/installation_instructions.md
+++ b/docs/src/installation_instructions.md
@@ -1,10 +1,24 @@
 # Installation
 
-To build the top-level project, first clone the repository, then instantiate:
+EnsembleKalmanProcesses.jl is a registered Julia package. You can install the latest version
+of EnsembleKalmanProcesses.jl through the built-in package manager. Press `]` in the Julia REPL
+command prompt and
+
+```julia
+julia> ]
+(v1.5) pkg> add EnsembleKalmanProcesses
+(v1.5) pkg> instantiate
+```
+
+This will install the latest tagged release of the package.
+
+### Cloning the repository
+
+Alternatively, you can clone the repository and then then instantiate:
 
 ```
-cd EnsembleKalmanProcesses.jl
-julia --project -e 'using Pkg; Pkg.instantiate()'
+> cd EnsembleKalmanProcesses.jl
+> julia --project -e 'using Pkg; Pkg.instantiate()'
 ```
 
 To test that the package is working:
@@ -18,8 +32,8 @@ To test that the package is working:
 Once the project is built, you can build the project documentation under the `docs/` sub-project:
 
 ```
-julia --project=docs/ -e 'using Pkg; Pkg.instantiate()'
-julia --project=docs/ docs/make.jl
+> julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()'
+> julia --project=docs/ docs/make.jl
 ```
 
 The locally rendered HTML documentation can be viewed at `docs/build/index.html`
--- a/docs/src/installation_instructions.md
+++ b/docs/src/installation_instructions.md
@@ -14,7 +14,7 @@ This will install the latest tagged release of the package.
 
 ### Cloning the repository
 
-Alternatively, you can clone the repository and then then instantiate:
+Alternatively, you can clone the repository and then instantiate:
 
 ```
 > cd EnsembleKalmanProcesses.jl
--- a/docs/src/installation_instructions.md
+++ b/docs/src/installation_instructions.md
@@ -30,7 +30,7 @@ julia> ]
 
 ### Cloning the repository
 
-If you are interested in having your hands dirty and modifying the code then, you can also
+If you are interested in getting your hands dirty and modifying the code then, you can also
 clone the repository and then instantiate, e.g.,
 
 ```
--- a/docs/src/installation_instructions.md
+++ b/docs/src/installation_instructions.md
@@ -17,7 +17,7 @@ This will install the latest tagged release of the package.
     
     ```julia
     julia> ]
-    (v1.5) pkg> add EnsembleKalmanProcesses#master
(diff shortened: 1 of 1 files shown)


glwagner (approved):
Seems good --- though I don't think we should recommend that people clone the repository at all. Only developers should clone the repository. Right @odunbar ?

navidcy:
Yeah, I just didn't want to change that as I was assuming it was there for a reason. Let me know and i can edit. Or anyone: feel free to edit ✍️

😀

odunbar:
If people want the examples they will need the repo. but in general to use the tool you do not need it.
Also, LGTM Thanks navidcy!

navidcy:
thanks @odunbar.
one question: what is this `bors` tests that is not passing?

glwagner:
> If people want the examples they will need the repo. but in general to use the tool you do not need it.
> Also, LGTM Thanks navidcy!

We've found with Oceanangians that a lot of users (not developers) end up copy/pasting examples from docs into the REPL to learn code. But it's true that we need to develop a way for people to download examples without cloning the repo (since repo cloning requires knowing git, which is a barrier for some).

navidcy:
@glwagner I added some clarifications in an attempt to address https://github.com/CliMA/EnsembleKalmanProcesses.jl/pull/40#issuecomment-903123776. Could you [have a look](https://clima.github.io/EnsembleKalmanProcesses.jl/previews/PR40/installation_instructions/)?

docs/src/installation_instructions.md
```diff
@@ -1,25 +1,61 @@
...
+You can run the tests via the package manager by:
+
+```julia
+julia> ]
+(v1.5) pkg> test EnsembleKalmanProcesses
+```
+
+### Cloning the repository
+
+If you are interested in having your hands dirty and modifying the code then, you can also
```
glwagner, line 33:
```suggestion
If you are interested in getting your hands dirty and modifying the code then, you can also
```

glwagner:
> @glwagner I added some clarifications in an attempt to address [#40 (comment)](https://github.com/CliMA/EnsembleKalmanProcesses.jl/pull/40#issuecomment-903123776). Could you [have a look](https://clima.github.io/EnsembleKalmanProcesses.jl/previews/PR40/installation_instructions/)?

Very nice clarification!

odunbar:
@glwagner I'm not sure encouraging copy-and-pasting from docs is really a solution here. Although yes with basic examples and with docs in the "literate" style this will run; you could just as easily give them the git commands to clone and run the example in a similar fashion with no understanding required - except perhaps they may need to create a git account to do it I guess?

@navidcy`bors` is the way that we perform merges into CliMA. Basically when the reviewers have approved the PR you send the message in `bor r+` and it will perform the merge.

glwagner:
> @glwagner I'm not sure encouraging copy-and-pasting from docs is really a solution here. Although yes with basic examples and with docs in the "literate" style this will run; you could just as easily give them the git commands to clone and run the example in a similar fashion with no understanding required - except perhaps they may need to create a git account to do it I guess?
> 
> @navidcy`bors` is the way that we perform merges into CliMA. Basically when the reviewers have approved the PR you send the message in `bor r+` and it will perform the merge.

Well, copy/paste into the REPL is the standard way to run julia code written in Documenter. We could develop another approach eventually but it might not be worth the time right now.

odunbar:
I agree to use the tool you wouldn't normally clone. I guess the simplest might then be to have an `Xexamples.jl` for each package `X` and just leave the source code and runtests with the original repo. But then i could see people wanting to use the examples as a form of testing too during development (not that we do this in an automated way) and this may cause other complications.

glwagner:
> I agree to use the tool you wouldn't normally clone. I guess the simplest might then be to have an `Xexamples.jl` for each package `X.jl` and just leave the source code and runtests with the original repo. But then i could see people wanting to use the examples as a form of testing too during development (not that we do this in an automated way) and this may cause other complications.
> 
> A few codes I have used in the past have contained an example experiments, and then you typically start by copying an experiment directory as a base.

Yeah, I think just providing links that download the examples in the documentation is a simple solution, eg

[Click to download this example](https://raw.githubusercontent.com/CliMA/Oceananigans.jl/master/examples/two_dimensional_turbulence.jl)

navidcy:
@glwagner and @odunbar, there are better solutions than "click to download this script", provided by Literate.jl.

Have a look e.g. [at GeophysicalFlows.jl docs](https://fourierflows.github.io/GeophysicalFlowsDocumentation/stable/generated/twodnavierstokes_decaying/). The example can be opened in nbviewer and downloaded as a Jupyter notebook. Literate.jl can also provide similar functionality to download the example as a .jl script without the markdown annotations.

Would something like that settle your discussion?

odunbar:
@glwagner Note, most of our examples is its own project (as they often have non-standard dependencies etc), I would be happy also with some kind of solution with a "click to download" perhaps from the docs, but it would need to be the whole example project not just the docs script.

@navidcy I think there is no one-size-fits-all solution for all as there are different use cases for the examples. 

A first user wants to run the examples and see the outputs, or in simple cases, play in the REPL. Your solution fits well here, or having documenter run REPL code in the docs.

A second user wants to develop the algorithmic tools will want to clone the repo itself, and also may wish to use the example collection to test the new tools (using a `dev` include of the repo in the examples). We may not have to cater too much to them, as they will need to be familiar with Julia/git here anyway.

A third user wants to create a new use case (using for example an existing code they have) for the existing tools. they likely want to download an existing similar example, create a new directory and modify it to handle the new model formulations/loss functions/data/priors/I-O of their setting, they may also need to use more flexible HPC interface with the tools (e.g. our Climate Machine example). These users likely wish to use the examples in a flexible way, as we can't envisage their application so will likely wish for more than notebooks can provide.

I quite like the "Download this example in a zip" button in the docs, and for simple cases the notebooks could help too!

odunbar:
I shall merge this

Thanks for the work and discussions :D

PR #41, opened 2021-08-20
Literate `loss_minimization.jl` example

navidcy:


navidcy:
I have liberated the example. 

It definitely wouldn't mind some more explanations. I may need some help on that because I'm not that familiar with the algorithm per se. But at least at this point the example appears in the docs with some animations that demonstrate the algorithms convergence.

navidcy:
A preview of how the example will appear in the docs is found at:
https://clima.github.io/EnsembleKalmanProcesses.jl/previews/PR41/literated/loss_minimization/

examples/LossMinimization/loss_minimization.jl
```diff
@@ -1,122 +1,201 @@
-using Distributions
-using LinearAlgebra
-using Random
-using Plots
-using EnsembleKalmanProcesses.EnsembleKalmanProcessModule
-using EnsembleKalmanProcesses.ParameterDistributionStorage
-
-# Seed for pseudo-random number generator for reproducibility
+# # Minimization Loss
```
glwagner, line 1: Is it "Minimization Loss" or "Loss Minimization" ???

examples/LossMinimization/loss_minimization.jl
```diff
@@ -1,122 +1,201 @@
...
+    Distributions,
+    LinearAlgebra,
+    Random,
+    Plots
+
+using
+    EnsembleKalmanProcesses.EnsembleKalmanProcessModule,
+    EnsembleKalmanProcesses.ParameterDistributionStorage
+
+# ## Loss function with single minimum
```
glwagner, line 15:
```suggestion
# ## Loss function with single minimum
#
# Here we minimize the loss function
# ```math
# G₁(u) = \|u - u_*\|,
# ```
# 
#  where ``u`` is a vector of parameters and ``u_*`` is a vector of optimal parameters
```

As a side note, it might make sense to use a different symbol than `u` since this package is intended for use with fluid dynamical problems...

glwagner:
Two comments are:

1. It probably makes sense to split this into two examples, one "one minimum" example, and one "two minimum example".
2. One technique to make the example render a bit more concisely in the docs is to combine lines, something like:

```julia
# We choose the number of observations and noise level,

n_obs = 1
noise_level = 1e-8
```

I understand this example much better now so I think this is a great improvement in clarity!

examples/LossMinimization/loss_minimization.jl
navidcy: I rephrased to "Minimization of Loss function"

navidcy:
> 1. It probably makes sense to split this into two examples, one "one minimum" example, and one "two minimum example".

You may be right..

navidcy:
Also, same question as in #40: what is this `bors` test and how can I make it happy?

odunbar:
Thanks for taking this on @navidcy , I confess I forgot to create these example docs so your efforts are much appreciated. If you wouldn't mind, I've added some small modifications to the examples too, as well as the docs.

Comments as I see them, 

1. This written this in a way that is very much unlike the style of the other examples. The content of the example itself is also very different. Perhaps we might think of calling this a "how-to" or "tutorial" or whatever the correct terminology is, rather than an example. 
1. LF: "Herem" -> Here
1. LF: "synthetic observations" -> synthetic observation
1. LF: Define the optimal parameters at the beginning,
1. LF: we write it as if there is observational noise on `y_obs` and yet we define `y_obs` without noise. (This is the reason why we get an unbiased result in the algorithm - because we use the noise as an algorithm stabilization here and not as actual observational noise). One thing we should do is call this noise `stabilization noise` rather than  `observational noise`, and e.g we should not call it `Gammay`, just `Gamma` or `Gamma_stab`. PS It also seems very small... maybe 1e-3 is sufficient and would make the method converge faster
1. LF:The prior should be in its own section please (as it does not define the loss function) 
1. LF: Parameterized distns why do we have 0., and -0., and why do we have sqrt(1). Maybe just call this  Normal(0,1), or if you like you can define a standard deviation/variance above
1. The `prior_mean` and `prior_cov` are not used?
1. C: Could you emphasize that the EKI is a choice of the available methods. And it is constructed by initializing with `Inversion()`
1. C: In the calibration loop could you mention somehow that we (i) obtain the parameters, (ii) calculate the loss function on the parameters (and concatenate) (iii) generate a new set of parameters using the model outputs.

1. LF2: Formula for G_2 as we did for G_1
1. LF2: why the assert on the minima? or rather, maybe we should output this and not just do it
1. LF2: Again new section for prior
1. LF2: Again the 0,-0 thing...
1. LF2: Again prior mean and cov not used

A small addition to the example which would be nice, would be to add a bias to the prior of u_1. i.e make the mean of the prior of the first distribution `Normal(-0.5,sqrt(2))` which will make convergence to the left hand minima more likely due to a good initial guess based on prior "intuition" rather than random chance based on the seed of the random number generator.

glwagner:
> Thanks for taking this on @navidcy , I confess I forgot to create these example docs so your efforts are much appreciated. If you wouldn't mind, I've added some small modifications to the examples too, as well as the docs.
> 
> Comments as I see them,
> 
> 1. This written this in a way that is very much unlike the style of the other examples. The content of the example itself is also very different. Perhaps we might think of calling this a "how-to" or "tutorial" or whatever the correct terminology is, rather than an example.
> 2. LF: "Herem" -> Here
> 3. LF: "synthetic observations" -> synthetic observation
> 4. LF: Define the optimal parameters at the beginning,
> 5. LF: we write it as if there is observational noise on `y_obs` and yet we define `y_obs` without noise. (This is the reason why we get an unbiased result in the algorithm - because we use the noise as an algorithm stabilization here and not as actual observational noise). One thing we should do is call this noise `stabilization noise` rather than  `observational noise`, and e.g we should not call it `Gammay`, just `Gamma` or `Gamma_stab`. PS It also seems very small... maybe 1e-3 is sufficient and would make the method converge faster
> 6. LF:The prior should be in its own section please (as it does not define the loss function)
> 7. LF: Parameterized distns why do we have 0., and -0., and why do we have sqrt(1). Maybe just call this  Normal(0,1), or if you like you can define a standard deviation/variance above
> 8. The `prior_mean` and `prior_cov` are not used?
> 9. C: Could you emphasize that the EKI is a choice of the available methods. And it is constructed by initializing with `Inversion()`
> 10. C: In the calibration loop could you mention somehow that we (i) obtain the parameters, (ii) calculate the loss function on the parameters (and concatenate) (iii) generate a new set of parameters using the model outputs.
> 11. LF2: Formula for G_2 as we did for G_1
> 12. LF2: why the assert on the minima? or rather, maybe we should output this and not just do it
> 13. LF2: Again new section for prior
> 14. LF2: Again the 0,-0 thing...
> 15. LF2: Again prior mean and cov not used
> 
> A small addition to the second example which would be nice, would be to add a bias to the prior of u_1. i.e make the mean of the prior of the first distribution `Normal(-0.5,sqrt(2))` which will make convergence to the left hand minima more likely due to a good initial guess based on prior "intuition" rather than random chance based on the seed of the random number generator.
> 
> Thanks again for your efforts, i think these changes shouldn't take long!

Might be helpful to use line comments for these!

https://docs.github.com/en/github/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/commenting-on-a-pull-request#adding-line-comments-to-a-pull-request

> This written this in a way that is very much unlike the style of the other examples. The content of the example itself is also very different. Perhaps we might think of calling this a "how-to" or "tutorial" or whatever the correct terminology is, rather than an example.

x-ref: https://github.com/CliMA/Oceananigans.jl/issues/1149

navidcy:
> This written this in a way that is very much unlike the style of the other examples. The content of the example itself is also very different. Perhaps we might think of calling this a "how-to" or "tutorial" or whatever the correct terminology is, rather than an example.

@odunbar, I confess that it was in my attempt to understand the algorithm and the example I sprinkled the comments. As it was before it was a bit cryptic/esoteric. :)

Perhaps it can be converted to a how-to or tutorial -- I'll let you decide on that.

I'll do the changes you suggested. Thanks! Then we can decide whether to call it something else. Or we can simply merge it as an example and convert it to something else later on.

navidcy:
@odunbar,


> 4. LF: Define the optimal parameters at the beginning,

Like where?

> 5. LF: we write it as if there is observational noise on `y_obs` and yet we define `y_obs` without noise. (This is the reason why we get an unbiased result in the algorithm - because we use the noise as an algorithm stabilization here and not as actual observational noise). One thing we should do is call this noise `stabilization noise` rather than  `observational noise`, and e.g we should not call it `Gammay`, just `Gamma` or `Gamma_stab`. PS It also seems very small... maybe 1e-3 is sufficient and would make the method converge faster

OK, I changed `Γ`->`Γ_stabilisation`. But feel free to add a remark on the code explaining the above? I also want to understand, so I don't yet feel comfortable doing so :)

> 7. LF: Parameterized distns why do we have 0., and -0., and why do we have sqrt(1). Maybe just call this  Normal(0,1), or if you like you can define a standard deviation/variance above

Haha... I was wondering the same. But I didn't wanna interfere!

> 11. LF2: Formula for G_2 as we did for G_1

Well, I avoided this because the way G₂ is defined it need to be explicitly defined via u components. It's kinda cumbersome and the code below is self-explanatory.

> 12. LF2: why the assert on the minima? or rather, maybe we should output this and not just do it

I don't understand what you mean here. Feel free to edit?
 
> A small addition to the second example which would be nice, would be to add a bias to the prior of u_1. i.e make the mean of the prior of the first distribution `Normal(-0.5,sqrt(2))` which will make convergence to the left hand minima more likely due to a good initial guess based on prior "intuition" rather than random chance based on the seed of the random number generator.

Done. But have a look and edit at your liking!

examples/LossMinimization/loss_minimization.jl
```diff
@@ -1,122 +1,217 @@
...
+# taken to be ``u_* = (-1, 1)``.
+#
+# We set the seed for pseudo-random number generator for reproducibility.
 rng_seed = 41
 Random.seed!(rng_seed)
+nothing # hide
 
-# Number of synthetic observations from G(u)
+# We choose the number of observations and noise level,
 n_obs = 1
```
glwagner, line 30:
```suggestion
N_observations = 1
```

examples/LossMinimization/loss_minimization.jl
```diff
@@ -1,122 +1,217 @@
...
-
-# Loss Function (unique minimum)
-function G(u)
-    return [sqrt((u[1]-1)^2 + (u[2]+1)^2)]
-end
+noise_level = 1e-3
+nothing # hide
+
+# Independent noise for synthetic observation:
+Γ_stabilisation = noise_level * Matrix(I, n_obs, n_obs)
```
glwagner, line 35:
```suggestion
Γ_stabilisation = noise_level * Matrix(I, N_observations, N_observations) 
```

examples/LossMinimization/loss_minimization.jl
```diff
@@ -1,122 +1,217 @@
...
-# Loss Function (unique minimum)
-function G(u)
-    return [sqrt((u[1]-1)^2 + (u[2]+1)^2)]
-end
+noise_level = 1e-3
+nothing # hide
+
+# Independent noise for synthetic observation:
+Γ_stabilisation = noise_level * Matrix(I, n_obs, n_obs) 
+noise = MvNormal(zeros(n_obs), Γ_stabilisation)
```
glwagner, line 36:
```suggestion
noise = MvNormal(zeros(N_observations), Γ_stabilisation)
```

examples/LossMinimization/loss_minimization.jl
```diff
@@ -1,122 +1,217 @@
...
-prior_mean = get_mean(prior)
-prior_cov = get_cov(prior)
 
-# Calibrate
-N_ens = 50  # number of ensemble members
-N_iter = 20 # number of EKI iterations
+# ### Calibration
+#
+# We choose the number of ensemble members and the number of EKI iterations¨
+N_ens  = 50
```
glwagner, line 62:
```suggestion
N_ensemble  = 50
```

examples/LossMinimization/loss_minimization.jl
```diff
@@ -1,122 +1,217 @@
...
+N_iter = 20
```
glwagner, line 63:
```suggestion
N_iterations = 20
```

examples/LossMinimization/loss_minimization.jl
```diff
@@ -1,122 +1,217 @@
...
+nothing # hide
+
+# With that in hand, we can construct our initial ensemble
 initial_ensemble = EnsembleKalmanProcessModule.construct_initial_ensemble(prior, N_ens;
                                                 rng_seed=rng_seed)
```
glwagner, line 68:
```suggestion
initial_ensemble =
    EnsembleKalmanProcessModule.construct_initial_ensemble(prior, N_ensemble; rng_seed=rng_seed)
```

examples/LossMinimization/loss_minimization.jl
```diff
@@ -1,122 +1,217 @@
...
 
-ekiobj = EnsembleKalmanProcessModule.EnsembleKalmanProcess(initial_ensemble,
-                    y_obs, Γy, Inversion())
-#
+# and the EKI. The EKI is a choice of the available methods and it is constructed by
+# initializing with `Inversion()` method.
+
+ekiobj = EnsembleKalmanProcessModule.EnsembleKalmanProcess(initial_ensemble, y_obs,
+                                                           Γ_stabilisation, Inversion())
```
glwagner, line 74:
```suggestion
ensemble_kalman_process =
    EnsembleKalmanProcessModule.EnsembleKalmanProcess(initial_ensemble, y_obs, Γ_stabilisation, Inversion())
```

examples/LossMinimization/loss_minimization.jl
```diff
@@ -1,122 +1,217 @@
...
+
+# Then we calibrate by *(i)* obtaining the parameters, *(ii)* calculate the loss function on
+# the parameters (and concatenate), and last *(iii)* generate a new set of parameters using
+# the model outputs:
 for i in 1:N_iter
     params_i = get_u_final(ekiobj)
```
glwagner, line 80:
```suggestion
for i in 1:N_iterations
    params_i = get_u_final(ensemble_kalman_process)
```

examples/LossMinimization/loss_minimization.jl
```diff
@@ -1,122 +1,217 @@
...
-    g_ens = hcat([G(params_i[:,i]) for i in 1:N_ens]...)
+    
+    g_ens = hcat([G₁(params_i[:, i]) for i in 1:N_ens]...)
```
glwagner, line 82:
```suggestion
    g_ens = hcat([G₁(params_i[:, i]) for i in 1:N_ensemble]...)
```

examples/LossMinimization/loss_minimization.jl
```diff
@@ -1,122 +1,217 @@
...
+    
     EnsembleKalmanProcessModule.update_ensemble!(ekiobj, g_ens)
```
glwagner, line 84:
```suggestion
    EnsembleKalmanProcessModule.update_ensemble!(ensemble_kalman_process, g_ens)
```

examples/LossMinimization/loss_minimization.jl
```diff
@@ -1,122 +1,217 @@
...
 end
 
+# and visualize the results:
 u_init = get_u_prior(ekiobj)
```
glwagner, line 88:
```suggestion
u_init = get_u_prior(ensemble_kalman_process)
```

examples/LossMinimization/loss_minimization.jl
```diff
@@ -1,122 +1,217 @@
...
 
-for i in 1:N_iter
-    u_i = get_u(ekiobj,i)
-    p = plot(u_i[1,:], u_i[2,:], seriestype=:scatter, xlims = extrema(u_init[1,:]), ylims = extrema(u_init[2,:]))
-    plot!([u_star[1]], xaxis="u1", yaxis="u2", seriestype="vline",
-        linestyle=:dash, linecolor=:red, label = false,
-        title = "EKI iteration = " * string(i)
+anim_unique_minimum = @animate for i in 1:N_iter
+    u_i = get_u(ekiobj, i)
```
glwagner, line 91:
```suggestion
anim_unique_minimum = @animate for i in 1:N_iterations
    u_i = get_u(ensemble_kalman_process, i)
```

examples/LossMinimization/loss_minimization.jl
glwagner: What exactly is `g_ens`? I can't puzzle it from the code.

examples/LossMinimization/loss_minimization.jl
```diff
@@ -1,122 +1,217 @@
...
+#
+# We then define the prior
+prior_distns = [Parameterized(Normal(0, 1)),
+                Parameterized(Normal(0, 1))]
+                
 constraints = [[no_constraint()], [no_constraint()]]
+
 prior_names = ["u1", "u2"]
+
 prior = ParameterDistribution(prior_distns, constraints, prior_names)
```
glwagner, line 57:
```suggestion
parameter_names = ["u1", "u2"]

prior = ParameterDistribution(prior_distns, constraints, parameter_names)
```

examples/LossMinimization/loss_minimization.jl
```diff
@@ -1,122 +1,217 @@
...
+y_obs  = G₁(u★)
 
-# Define Prior
-prior_distns = [Parameterized(Normal(0., sqrt(1))),
-                Parameterized(Normal(-0., sqrt(1)))]
+# ### Prior distributions
+#
+# We then define the prior
+prior_distns = [Parameterized(Normal(0, 1)),
+                Parameterized(Normal(0, 1))]
```
glwagner, line 51:
```suggestion
prior_distributions = [Parameterized(Normal(0, 1)), Parameterized(Normal(0, 1))]
```

examples/LossMinimization/loss_minimization.jl
```diff
@@ -1,122 +1,217 @@
-using Distributions
-using LinearAlgebra
-using Random
-using Plots
-using EnsembleKalmanProcesses.EnsembleKalmanProcessModule
-using EnsembleKalmanProcesses.ParameterDistributionStorage
-
-# Seed for pseudo-random number generator for reproducibility
+# # Minimization of Loss function
```
glwagner, line 1:
```suggestion
# # Minimization of simple loss function with unique minimum
```

glwagner:
All applications of EKI are examples of "loss minimization", right? What's a better name for this example?

examples/LossMinimization/loss_minimization.jl
glwagner: I don't think `noise` is used anywhere in this file; therefore these lines are irrelevant to the example. @odunbar is this a typo / mistake?

navidcy:
thanks @glwagner!

navidcy:
nice @odunbar!

PR #47, opened 2021-09-02
Clarifies notation and units in Ensemble Kalman Inversion documentation

### LINKED ISSUE #46: Clarify notation and units in Ensemble Kalman Inversion documentation?
A few clarifications, definitions of symbols, and perhaps units in the [docs on Ensemble Kalman Inversion](https://clima.github.io/EnsembleKalmanProcesses.jl/dev/ensemble_kalman_inversion/#Ensemble-Kalman-Inversion) might improve the docs-reading experience. The docs now read

image

For example, the meaning of "curly N"

image

it's parameters, or the units of the covariance matrix are not specified. This leads to questions about whether the parameter update formula has the correct dimensions. The inner produce <., .> is also not defined.

navidcy:
Attempts to resolve #46.

### DIFF
--- a/docs/src/ensemble_kalman_inversion.md
+++ b/docs/src/ensemble_kalman_inversion.md
@@ -32,7 +32,7 @@ across ensemble members,
 ```
 
 and angle brackets denote the Euclidean inner product. By multiplying with ``\Gamma_y^{-1}``
-(which has dimensions ``y^2``) we render the inner product non-dimensional.
+we render the inner product non-dimensional.
 
 The EKI algorithm is considered converged when the ensemble achieves sufficient consensus/collapse
 in parameter space. The final estimate ``\bar{\theta}_{N_{it}}`` is taken to be the ensemble
--- a/docs/src/ensemble_kalman_inversion.md
+++ b/docs/src/ensemble_kalman_inversion.md
@@ -1,37 +1,48 @@
 # Ensemble Kalman Inversion
 
-One of the ensemble Kalman processes implemented in `EnsembleKalmanProcesses.jl` is the ensemble Kalman inversion ([Iglesias et al, 2013](http://dx.doi.org/10.1088/0266-5611/29/4/045001)). The ensemble Kalman inversion (EKI) is a derivative-free ensemble optimization method that seeks to find the optimal parameters $\theta \in \mathbb{R}^p$ in the inverse problem
-
-\\[ y = \mathcal{G}(\theta) + \eta, \\]
-
-where $\mathcal{G}$ denotes the forward map, $y \in \mathbb{R}^d$ is the vector of observations and $\eta \sim \mathcal{N}(0, \Gamma_y)$ is additive Gaussian observational noise. Note that $p$ is the size of the parameter vector $\theta$ and $d$ is taken to be the size of the observation vector $y$. The EKI update equation for parameter vector $\theta^{(j)}$ of ensemble member $j$ is
+One of the ensemble Kalman processes implemented in `EnsembleKalmanProcesses.jl` is the ensemble Kalman inversion ([Iglesias et al, 2013](http://dx.doi.org/10.1088/0266-5611/29/4/045001)). The ensemble Kalman inversion (EKI) is a derivative-free ensemble optimization method that seeks to find the optimal parameters ``\theta \in \mathbb{R}^p`` in the inverse problem
 
 ```math
-\theta_{n+1}^{(j)} = \theta_{n}^{(j)} - \dfrac{\Delta t_n}{J}\sum_{k=1}^J\langle \mathcal{G}(\theta_n^{(k)}) - \bar{\mathcal{G}}_n, \Gamma_y^{-1}(\mathcal{G}(\theta_n^{(j)}) - y) \rangle \theta_{n}^{(k)},
+y = \mathcal{G}(\theta) + \eta,
 ```
 
-where the subscript $n=1, \dots, N_{it}$ indicates the iteration, $J$ is the number of members in the ensemble and $\bar{\mathcal{G}}_n$ is the mean value of $\mathcal{G}(\theta)$ across ensemble members,
+where ``\mathcal{G}`` denotes the forward map, ``y \in \mathbb{R}^d`` is the vector of observations
+and ``\eta \sim \mathcal{N}(0, \Gamma_y)`` is additive Gaussian observational noise. Here, 
+``\mathcal{N}(0, \Gamma_y)`` denotes a normal distribution with mean 0 and variance ``\Gamma_y``.
+Note that ``p`` is the size of the parameter vector ``\theta`` and ``d`` is taken to be the size of the observation vector ``y``.
+
+The EKI update equation for parameter vector ``\theta^{(j)}`` of ensemble member ``j`` is
 
 ```math
-\bar{\mathcal{G}}_n = \dfrac{1}{J}\sum_{k=1}^J\mathcal{G}(\theta_n^{(k)}).
+\theta_{n+1}^{(j)} = \theta_{n}^{(j)} - \dfrac{\Delta t_n}{J}\sum_{k=1}^J \left \langle \frac{\mathcal{G}(\theta_n^{(k)}) - \bar{\mathcal{G}}_n}{\sqrt{\Gamma_y}} \, , \, \frac{\mathcal{G}(\theta_n^{(j)}) - y}{\sqrt{\Gamma_y}} \right \rangle \theta_{n}^{(k)},
 ```
 
-The EKI algorithm is considered converged when the ensemble achieves sufficient consensus/collapse in parameter space. The final estimate $\bar{\theta}_{N_{it}}$ is taken to be the ensemble mean at the final iteration,
+where the subscript ``n=1, \dots, N_{it}`` indicates the iteration, ``J`` is the number of members in the ensemble and ``\bar{\mathcal{G}}_n`` is the mean value of ``\mathcal{G}(\theta)`` across ensemble members,
+
+```math
+\bar{\mathcal{G}}_n = \dfrac{1}{J}\sum_{k=1}^J\mathcal{G}(\theta_n^{(k)}),
+```
+
+and angle brackets denote a Euclidean inner product. By dividing with ``\sqrt{\Gamma_y}`` we
+render the elements inside the inner product non-dimensional.
+
+The EKI algorithm is considered converged when the ensemble achieves sufficient consensus/collapse in parameter space. The final estimate ``\bar{\theta}_{N_{it}}`` is taken to be the ensemble mean at the final iteration,
 
 ```math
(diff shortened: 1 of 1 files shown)


navidcy:
@glwagner and @adelinehillier, how does it read now? Feel free to edit at your liking.

glwagner:
@navidcy do we have docs previews?

glwagner (approved):

glwagner:
Thanks @navidcy, I appreciate the extra verbosity.

navidcy:
> Is Gamma_y a scalar variance or a covariance matrix (eg curly N is the multivariate Gaussian)... ?

Hm, didn't think of that. 
Still Γ is positive definite and thus the inverse square root makes sense, right?

navidcy:
Even if it's a matrix, it should be a positive definite one. The inverse square root is then, nothing else, from the eigen decomposition of Γ but with the inverse square root of the eigenvalues of Γ in the diagonal matrix. Right?

navidcy:
But perhaps it's simpler if we go back to Γ^{-1}. :)

glwagner:
> Even if it's a matrix, it should be a positive definite one. The inverse square root is then, nothing else, from the eigen decomposition of Γ but with the inverse square root of the eigenvalues of Γ in the diagonal matrix. Right?

Ah ok.

> But perhaps it's simpler if we go back to Γ^{-1}. :)

Your call.

navidcy:
> > Even if it's a matrix, it should be a positive definite one. The inverse square root is then, nothing else, from the eigen decomposition of Γ but with the inverse square root of the eigenvalues of Γ in the diagonal matrix. Right?
> 
> Ah, I see.

Well, now that I'm thinking about it, same questions arise if Γ is a matrix and we write Γ^{-1}. Does the inverse exist and is it well-defined? So I vote for keeping the symmetric version and adding a clarification, I'll do that. :)

glwagner:
> > > Even if it's a matrix, it should be a positive definite one. The inverse square root is then, nothing else, from the eigen decomposition of Γ but with the inverse square root of the eigenvalues of Γ in the diagonal matrix. Right?
> > 
> > 
> > Ah, I see.
> 
> Well, now that I'm thinking about it, same questions arise if Γ is a matrix and we write Γ^{-1}. Does the inverse exist and is it well-defined? So I vote for keeping the symmetric version and adding a clarification, I'll do that. :)

Okay. Also need to clarify that Gamma_y is a covariance matrix rather than variance and N is the multivariate Gaussian distribution. It could be nice to remark on the meaning of Gamma_y too (represents uncertainty, including correlated uncertainty, in the observations?)

navidcy:
@glwagner have a look at 46b0969 and feel free to rephrase to make it even clearer! :)

docs/src/ensemble_kalman_inversion.md
```diff
@@ -1,37 +1,56 @@
...
+One of the ensemble Kalman processes implemented in `EnsembleKalmanProcesses.jl` is the ensemble Kalman inversion ([Iglesias et al, 2013](http://dx.doi.org/10.1088/0266-5611/29/4/045001)). The ensemble Kalman inversion (EKI) is a derivative-free ensemble optimization method that seeks to find the optimal parameters ``\theta \in \mathbb{R}^p`` in the inverse problem
 
-\\[ y = \mathcal{G}(\theta) + \eta, \\]
+```math
+y = \mathcal{G}(\theta) + \eta,
+```
+
+where ``\mathcal{G}`` denotes the forward map, ``y \in \mathbb{R}^d`` is the vector of observations
+and ``\eta  \in \mathbb{R}^d`` is additive Gaussian observational noise. Here, we take 
+``\eta \sim \mathcal{N}(0, \Gamma_y)`` from a multivariate ``d``-dimensional normal distribution
```
odunbar, line 11: You say eta is Gaussian twice. Also it is more common to say d-dimensional multivariate normal rather than multivariate ``d``-dimensional normal.

docs/src/ensemble_kalman_inversion.md
```diff
@@ -1,37 +1,56 @@
...
+with zero mean and covariance matrix ``\Gamma_y``.  This noise structure aims to represent the
+uncertainty, including correlated uncertainty, in the observations. Note that ``p`` is the
+size of the parameter vector ``\theta`` and ``d`` is taken to be the size of the observation
+vector ``y``.
```
odunbar, line 15: repetition of definition p and d? perhaps just write this out when y and theta are defined?

"uncertainty, including correlated uncertainty, in the observations" -> "correlations between observations"

docs/src/ensemble_kalman_inversion.md
```diff
@@ -1,37 +1,56 @@
...
 
-where $\mathcal{G}$ denotes the forward map, $y \in \mathbb{R}^d$ is the vector of observations and $\eta \sim \mathcal{N}(0, \Gamma_y)$ is additive Gaussian observational noise. Note that $p$ is the size of the parameter vector $\theta$ and $d$ is taken to be the size of the observation vector $y$. The EKI update equation for parameter vector $\theta^{(j)}$ of ensemble member $j$ is
+The parameter vector of the ``j``-th ensemble member at the ``n``-th iteration is ``\theta^{(j)}_n``. The EKI update equation for parameter vector ``\theta^{(j)}`` is
 
 ```math
-\theta_{n+1}^{(j)} = \theta_{n}^{(j)} - \dfrac{\Delta t_n}{J}\sum_{k=1}^J\langle \mathcal{G}(\theta_n^{(k)}) - \bar{\mathcal{G}}_n, \Gamma_y^{-1}(\mathcal{G}(\theta_n^{(j)}) - y) \rangle \theta_{n}^{(k)},
+\theta_{n+1}^{(j)} = \theta_{n}^{(j)} - \dfrac{\Delta t_n}{J}\sum_{k=1}^J \left \langle \frac{\mathcal{G}(\theta_n^{(k)}) - \bar{\mathcal{G}}_n}{\sqrt{\Gamma_y}} \, , \, \frac{\mathcal{G}(\theta_n^{(j)}) - y}{\sqrt{\Gamma_y}} \right \rangle \theta_{n}^{(k)} ,
```
odunbar, line 20: We do not use the square root notation of the matrix in implementation (this implies one needs to factorize, compute the eigenvalues, and square root them, which we don't do), I would keep it as Gamma^{-1} on the second term - this is also in keeping with all statistical methods papers which are based on Gaussian methods

docs/src/ensemble_kalman_inversion.md
```diff
@@ -1,37 +1,56 @@
...
 ```math
-\bar{\mathcal{G}}_n = \dfrac{1}{J}\sum_{k=1}^J\mathcal{G}(\theta_n^{(k)}).
+\bar{\mathcal{G}}_n = \dfrac{1}{J}\sum_{k=1}^J\mathcal{G}(\theta_n^{(k)}) ,
 ```
 
-The EKI algorithm is considered converged when the ensemble achieves sufficient consensus/collapse in parameter space. The final estimate $\bar{\theta}_{N_{it}}$ is taken to be the ensemble mean at the final iteration,
+and angle brackets denote the Euclidean inner product. The covariance matrix ``\Gamma_y`` is
+positive-definite and, therefore, raising it to the power ``-1/2`` is meaningful. By normalizing
+with ``\Gamma_y^{-1/2}`` we render the elements inside the inner product non-dimensional.
+
```
odunbar, line 34: Remove the positive definiteness and -1/2 stuff. We can keep the mention of nondimensionalization, I think this is relevant to user understanding.

docs/src/ensemble_kalman_inversion.md
```diff
@@ -64,18 +83,19 @@ A call to the inversion algorithm can be performed with the `update_ensemble!` f
 
 A typical use of the `update_ensemble!` function given the ensemble Kalman inversion object `ekiobj`, the dynamical model `Ψ` and the observation map `H` is
 ```julia
-N_iter = 20 # Number of steps of the algorithm
+N_iterations = 20 # Number of steps of the algorithm
```
odunbar, line 86: We have a glossary of terms for consistency amongst documentation and code for the inverse problem examples.

https://clima.github.io/EnsembleKalmanProcesses.jl/dev/glossary/

Though it is fine to change `N_iter` to `N_iteration` if you believe it is needed, let's do it in a separate PR, as it needs to be changed everywhere.

odunbar (commented):
Tahnks! A few small comments. 

RE notation - we have a[ glossary  ](https://clima.github.io/EnsembleKalmanProcesses.jl/dev/glossary/) which links notation in code and notes. We have tried to be consistent here, so changing these will require its own PR. as things will need to change everywhere.

RE the formulas, the notation we use matches papers that describe the algorithms as well as how they are implemented, so i think we should leave these.

navidcy:
> RE notation - we have a[ glossary ](https://clima.github.io/EnsembleKalmanProcesses.jl/dev/glossary/) which links notation in code docs and the application examples. We have tried to be consistent here, so changing these will require its own PR. as things will need to change everywhere.

Perhaps raise this as an issue (if it's not already there)?

docs/src/ensemble_kalman_inversion.md
navidcy: It's not *needed*, but what is really needed? We can get philosophical here. It's just much easier to read code if it's verbose and here 4-5 characters make the semi-incomprehensible `N_iter` to read almost English. :)

navidcy:
preview @ https://clima.github.io/EnsembleKalmanProcesses.jl/previews/PR47/ensemble_kalman_inversion/

navidcy:
@odunbar I did everything, do you want to have another look?

docs/src/ensemble_kalman_inversion.md
```diff
@@ -1,37 +1,58 @@
...
+across ensemble members,
 
 ```math
-\bar{\mathcal{G}}_n = \dfrac{1}{J}\sum_{k=1}^J\mathcal{G}(\theta_n^{(k)}).
+\bar{\mathcal{G}}_n = \dfrac{1}{J}\sum_{k=1}^J\mathcal{G}(\theta_n^{(k)}) ,
 ```
 
-The EKI algorithm is considered converged when the ensemble achieves sufficient consensus/collapse in parameter space. The final estimate $\bar{\theta}_{N_{it}}$ is taken to be the ensemble mean at the final iteration,
+and angle brackets denote the Euclidean inner product. By multiplying with ``\Gamma_y^{-1}``
+(which has dimensions ``y^2``) we render the inner product non-dimensional.
```
odunbar, line 35: I think you mean ``d \times d`` rather than ``y^2``

odunbar (approved):
I'm happy except for the typo, LGTM, thanks Navid!

PS you can view the docs by going the the bottom of the thread, and  click `Details` in the `documenter/deploy — Documentation build succeeded ` section

docs/src/ensemble_kalman_inversion.md
navidcy: I actually meant "units of y^2". Thanks.

navidcy: But I'll just remove the parenthetical remark.

PR #56, opened 2021-09-27
unpin Distributions version

costachris:
The Distributions package pinned previously was incompatible with the new TC.jl environment.


haakon-e:
Is there any reason this package should be pinned, @odunbar?

odunbar:
I believe we required some kind of compat for the version release. So I would rather you didn't delete it outright.
What would be more useful, is if you can expand the allowable range in compat to be compatible with TC. 

Could run the unit tests offline to make sure there aren't issues too? (with the compatible version of Distributions?)

costachris:
@odunbar Okay I used a range instead, which fixes our problem. All tests in `test/runtests.jl` pass locally.

ilopezgp (approved):

PR #60, opened 2021-10-02
Remove unnecessary calls and inputs in UKI.

ilopezgp:
This PR cleans up some unused calls and inputs in the Unscented Kalman constructor. Behavior is unchanged, except for the call to `Unscented`, which now has one argument less.


charleskawczynski (approved):
LGTM

PR #61, opened 2021-10-02
The great formatting.

ilopezgp:
Let's use the formatting that makes other CliMA packages look so nice.

### DIFF
--- a/.dev/Project.toml
+++ b/.dev/Project.toml
@@ -1,0 +1,5 @@
+[deps]
+JuliaFormatter = "98e50ef6-434e-11e9-1051-2b60c6c9e899"
+
+[compat]
+JuliaFormatter = "0.3"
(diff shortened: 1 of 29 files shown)


charleskawczynski (commented):
Looks good to me. We probably want to add a `"format"` entry to the [bors toml file](https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/main/bors.toml#L3).

We could also use bors and add branch protection, it looks like.

ilopezgp:
I added the `format` entry, but I am not sure how to add branch protection.

charleskawczynski:
I can do that

charleskawczynski:
Let's first try merging with bors, to make sure that we're passing that way.

charleskawczynski (approved):

PR #66, opened 2021-10-10
Create GitHub action that cleans up the Docs preview versions from PRs

navidcy:
based on https://github.com/CliMA/TimeMachine.jl/blob/master/.github/workflows/DocCleanup.yml

### DIFF
--- a/.github/workflows/DocPreviewCleanup.yml
+++ b/.github/workflows/DocPreviewCleanup.yml
@@ -1,0 +1,26 @@
+name: Doc Preview Cleanup
+
+on:
+  pull_request:
+    types: [closed]
+
+jobs:
+  doc-preview-cleanup:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout gh-pages branch
+        uses: actions/checkout@v2
+        with:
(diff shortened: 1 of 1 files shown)


charleskawczynski:
The GHA in https://github.com/CliMA/TimeMachine.jl/blob/master/.github/workflows/DocCleanup.yml seems to look a bit different. In particular, the `if [ -d "previews/PR$PRNUM" ]; then` in TimeMachine.jl seems like this happens conditionally, which doesn't seem to be the case in this PR. Is this an updated recommended version from somewhere?

navidcy:
> The GHA in https://github.com/CliMA/TimeMachine.jl/blob/master/.github/workflows/DocCleanup.yml seems to look a bit different. In particular, the `if [ -d "previews/PR$PRNUM" ]; then` in TimeMachine.jl seems like this happens conditionally, which doesn't seem to be the case in this PR. Is this an updated recommended version from somewhere?

I got this from Documenter.jl Docs --> https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#gh-pages-Branch

PR #70, opened 2021-11-03
[WIP] Add implementation of sparse EKI

jinlong83:
This PR adds the option of sparse EKI to Ensemble Kalman Inversion process:

- Added the option of sparse EKI to `EnsembleKalmanInversion.jl`
- Added an example (`LossMinimization/loss_minimization_sparse_eki.jl`)


odunbar:
Hi Jinlong, the way we have been adding different methods is by keeping them separated rather than using flags. As the sparsity preserving EKI is sufficiently different from EKI maybe we could do it in this case. use the `Sampler()` type as a guide (in `EnsembleKalmanSampler.jl`)

Could you create a new file `SparseEnsembleKalmanInversion.jl` and inside you can add 
- a Type called `SparseInversion` which takes in all the "Sparse method only" parameters
- an `update_ensemble(ekp::EnsembleKalmanProcess{FT, IT, SparseInversion{FT}}, ...)` method where you put in the sparse update?

We can call to chat more about this

odunbar:
Note I got this to run by first 
```
conda install -c conda-forge cvxopt
```
and then in the project base
```
julia --project
]
add CVXOPT
```
then in the example directory
```
julia --project loss_minimization_sparse.jl
```

ilopezgp:
Is there a way to implemented the Sparse EKI without `PyCall`? Python dependencies are a headache, and should be avoided if we can.

jinlong83:
> Is there a way to implement the Sparse EKI without `PyCall`? Python dependencies are a headache, and should be avoided if we can. This might be useful: https://jump.dev/Convex.jl/stable/

I have updated the implementation to avoid `PyCall`. It seems that `Convex.jl` is slightly less stable than `CVXOPT`, so I also updated the default settings of sparse EKI parameters in the example.

Project.toml
```diff
@@ -4,6 +4,7 @@ authors = ["CLIMA contributors <clima-software@caltech.edu>"]
 version = "0.1.1"
 
 [deps]
+CVXOPT = "1a0d837d-f7ba-5e72-a8f8-b8273492ab4e"
```
jakebolewski, line 7: Need to add Convex to the depdencies

jinlong83: I just updated the dependencies and locally ran through the tests.

odunbar:
I forgot to mention the unit tests.

Could you transfer the sparse example you created, and place it in the file `test/EnsembleKalmanProcessModule/runtests.jl`   

inside you then add some`@test` statements - which are like fancy asserts. You can use the the `EnsembleKalmanInversion` testing done in the file as aguide, (and maybe add something to test the sparsity property?)

To checkthe testing offline you go to the base and do 
```
julia --project
using Test
include("test/runtests.jl")
```

odunbar (approved):
Excellent addition. LGTM

Please add a check for the L^1 norm in the sparse eki loop.

If you could create a docs PR soon, that adds a page linking to your paper / what the sparsity means in this context etc.  that would be awesome.

test/EnsembleKalmanProcessModule/runtests.jl
```diff
@@ -211,4 +211,137 @@ using EnsembleKalmanProcesses.ParameterDistributionStorage
...
+    ###
+
+    N_ens = 20 # number of ensemble members
+    N_iter = 5 # number of EKI iterations
+    initial_ensemble = EnsembleKalmanProcessModule.construct_initial_ensemble(prior, N_ens; rng_seed = rng_seed)
+    @test size(initial_ensemble) == (n_par, N_ens)
+
+    # Sparse EKI parameters
+    γ = 1.0
+    threshold_eki = false
```
jakebolewski, line 249: you could write these as a loop ex.

for (threshold_eki, threshold_value) in ((false, 1e-2), (true, 1e2)
    ....
    
end

src/SparseEnsembleKalmanInversion.jl
```diff
@@ -0,0 +1,153 @@
...
+using Convex, SCS
+using SparseArrays
+
+"""
+    SparseInversion <: Process
+
+A sparse ensemble Kalman Inversion process
+"""
+struct SparseInversion{FT <: AbstractFloat, IT <: Int} <: Process
+    ""
```
ilopezgp, line 11: Please add a docstring describing each of these struct elements, like in `UnscentedKalmanInversion.jl`.

src/SparseEnsembleKalmanInversion.jl
```diff
@@ -0,0 +1,153 @@
...
+    threshold_value::FT
+    ""
+    reg::FT
+    ""
+    uc_idx::Vector{IT}
+end
+
+"""
+    sparse_qp(ekp::EnsembleKalmanProcess{FT, IT, SparseInversion{FT,IT}}, v_j::Vector{FT}, cov_vv_inv::Array{FT, 2}, H_u::SparseArrays.SparseMatrixCSC{FT}, H_g::SparseArrays.SparseMatrixCSC{FT}, y_j::Vector{FT}, γ::FT; H_uc::SparseArrays.SparseMatrixCSC{FT} = H_u) where {FT, IT}
```
ilopezgp, line 24: These are not formatted automatically, but it is better to follow the actual function format below for readability, this line is too long.

src/SparseEnsembleKalmanInversion.jl
```diff
@@ -0,0 +1,153 @@
...
+Solving quadratic programming problem with sparsity constraint.
+"""
+function sparse_qp(
+    ekp::EnsembleKalmanProcess{FT, IT, SparseInversion{FT, IT}},
+    v_j::Vector{FT},
+    cov_vv_inv::Array{FT, 2},
+    H_u::SparseArrays.SparseMatrixCSC{FT},
+    H_g::SparseArrays.SparseMatrixCSC{FT},
+    y_j::Vector{FT},
+    γ::FT;
```
ilopezgp, line 35: Is this the γ that is part of the SparseInversion? If so, can you retrieve it from `ekp`?

src/SparseEnsembleKalmanInversion.jl
```diff
@@ -0,0 +1,153 @@
...
+    H_uc::SparseArrays.SparseMatrixCSC{FT} = H_u,
+) where {FT, IT}
+
+    P = H_g' * inv(ekp.obs_noise_cov) * H_g + cov_vv_inv
```
ilopezgp, line 39: Please avoid inverse computations when solving linear systems, do

`P = H_g' * (ekp.obs_noise_cov \ H_g) + cov_vv_inv`

instead.

src/SparseEnsembleKalmanInversion.jl
```diff
@@ -0,0 +1,153 @@
...
+    P = 0.5 * (P + P')
+    q = -(cov_vv_inv * v_j + H_g' * inv(ekp.obs_noise_cov) * y_j)
```
ilopezgp, line 41:
```suggestion
    q = -(cov_vv_inv * v_j + H_g' * ( ekp.obs_noise_cov \ y_j))
```

src/SparseEnsembleKalmanInversion.jl
```diff
@@ -0,0 +1,153 @@
...
+    # Sparse EKI
+    cov_vv = vcat(hcat(cov_uu, cov_ug), hcat(cov_ug', cov_gg))
+    H_u = hcat(1.0 * I(size(u)[1]), fill(FT(0), size(u)[1], size(g)[1]))
+    H_g = hcat(fill(FT(0), size(g)[1], size(u)[1]), 1.0 * I(size(g)[1]))
+
+    H_uc = H_u[ekp.process.uc_idx, :]
+
+    Threads.@threads for j in 1:(ekp.N_ens)
+        # Solve a quadratic programming problem
+        u[:, j] = sparse_qp(ekp, v[j, :], inv(cov_vv), H_u, H_g, y[:, j], ekp.process.γ, H_uc = H_uc)
```
ilopezgp, line 126: Can you pass `cov_vv` instead of `inv(cov_vv)` and solve the linear system in `sparse_qp`?

test/EnsembleKalmanProcessModule/runtests.jl
jinlong83: The for loop has been implemented as suggested.

src/SparseEnsembleKalmanInversion.jl
jinlong83: I have added a doctoring description.

src/SparseEnsembleKalmanInversion.jl
jinlong83: I have adjusted the format as suggested.

src/SparseEnsembleKalmanInversion.jl
jinlong83: Yes it is. I have changed the implementation to retrieve it from `ekp`.

src/SparseEnsembleKalmanInversion.jl
jinlong83: I have updated the implementations to avoid directly calling matrix inverse.

src/SparseEnsembleKalmanInversion.jl
```diff
@@ -0,0 +1,168 @@
...
+    cov_vv = vcat(hcat(cov_uu, cov_ug), hcat(cov_ug', cov_gg))
+    H_u = hcat(1.0 * I(size(u)[1]), fill(FT(0), size(u)[1], size(g)[1]))
+    H_g = hcat(fill(FT(0), size(g)[1], size(u)[1]), 1.0 * I(size(g)[1]))
+
+    H_uc = H_u[ekp.process.uc_idx, :]
+
+    cov_vv_inv = cov_vv \ (1.0 * I(size(cov_vv)[1]))
+
+    # Loop over ensemble members to impose sparsity
+    Threads.@threads for j in 1:(ekp.N_ens)
```
ilopezgp, line 139: The use of `Threads.@threads` is the only thing I am not sure about. Do we want to have this here, @jakebolewski ?

ilopezgp (approved):
After my `Threads.@threads` question has been addressed by someone more experienced with Julia, LGTM.

src/SparseEnsembleKalmanInversion.jl
jakebolewski: @ilopezgp it should be fine as long as it's not calling into any thread-unsafe library (like the conic solver, one would have to check if that is thread safe). 

IIRC I asked @jinlong83 about this a while back and it looked like there were issues with printing but I believe that is disabled

jakebolewski: It looks like SCS is mostly thread safe? https://github.com/cvxgrp/scs/pull/48

PR #72, opened 2021-11-15
added a copy

### LINKED ISSUE #71: I'm worried that we don't do enough copies.
https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/249133ceddf302ab5999343895aae189b04202c1/src/DataStorage.jl#L24

I think when we put stuff in storage, or retrieve it, I have concerns that none of the reference links are being broken. Are there some safety measures we can do to help here to enforce copies or something?

The last thing i caught, was when i thought there was an issue making a gif, but it actually resulted from the lack of a copy here:
https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/249133ceddf302ab5999343895aae189b04202c1/src/EnsembleKalmanInversion.jl#L66

in EKI leading to every stored value being changed into the final one.

odunbar:
## Purpose
Resolves #71, when we load data from data storage we should always copy, as it is history and should not be overwritable

## Example
I was recently stung by the old mutability vs assignment:
```julia 
x=ones(10)
f() = return x
```
then 
```julia 
u1 =  f()
u1 += ones(10)
# u1 = 2*ones(10)
# x = ones(10)
```
and
```julia 
u2 = f()
u2 = 2*ones(10)
# u2 = 2*ones(10)
# x = ones(10)
```
But(!)
```julia 
u3 = f()
u3[:] = 2*ones(10)
# u2 = 2*ones(10)
# x = 2*ones(10)
```
let's not rely on people like me to remember this in e.g. the getters from storage `get_u`,`get_u_final`,`get_u_prior` etc.
 A copy in the `get_data` method of the `DataStorage` class resolves this
```julia 
x=ones(10)
g() = return copy(x)
u4 = g()
u4[:] = 2*ones(10)
# u4 = 2*ones(10)
# x = ones(10)
```

### DIFF
--- a/src/DataStorage.jl
+++ b/src/DataStorage.jl
@@ -97,7 +97,7 @@ function size(pdc::PairedDataContainer, idx::IT) where {IT <: Integer}
 end
 
 function get_data(dc::DataContainer)
-    return copy(dc.stored_data)
+    return deepcopy(dc.stored_data)
 end
 function get_data(pdc::PairedDataContainer)
     return get_inputs(pdc), get_outputs(pdc)
--- a/src/DataStorage.jl
+++ b/src/DataStorage.jl
@@ -97,7 +97,7 @@ function size(pdc::PairedDataContainer, idx::IT) where {IT <: Integer}
 end
 
 function get_data(dc::DataContainer)
-    return dc.stored_data
+    return copy(dc.stored_data)
 end
 function get_data(pdc::PairedDataContainer)
     return get_inputs(pdc), get_outputs(pdc)


ilopezgp (approved):

jakebolewski:
You probably want to use `deepcopy` otherwise LGTM

bielim:
Looks good to me!
For some reason this instance of the problem never bit me, but with python I fell into that same trap a couple of times. And I had to learn that the trap also exists in languages like C++, where you'd think that because everything is more explicit you might be less prone to accidentally changing the values of variables you don't want to change, for example:
```
#include <algorithm>  
#include <vector>

#include <cassert>

using namespace std;

int main() {

  float const pi  { 3.1415f };

  // x = [1.,1.,1.]
  auto x = vector<float>(3, 1.f);

  // A C++ reference
  auto& u = x;

  // When you assign to a reference in C++ it is as if you assigned to the 
  // original value (because the reference has value semantics!), so in
  // contrast to Julia (and python), the following re-assignment will
  // actually change x as well:
  u = std::vector<float>(3, pi);

  // So x now also became pi
  assert(all_of(x.cbegin(), x.cend(), [&](auto v){ return v == pi; }));

  return 0;
}
```
Yay :-)

bielim (approved):

PR #75, opened 2021-12-03
Run tests and build docs with julia v1.6

navidcy:


navidcy:
Why the Docs GitHub action is not even triggered? Weird!

PR #74, opened 2021-12-03
Better formatting for some docstrings

navidcy:

PR #68, opened 2021-10-11
Docs/Prior Distributions section enhancements

navidcy:
This PR increases the line width in the Docs/Prior Distributions section. Also improves the code formatting and simplifies code a bit in the examples in this section.

PR #77, opened 2021-12-03
Decouple initial conditions in UKI from prior for regularization.

ilopezgp:
The initialization of the mean in UKI need not be the prior mean. In some situations, we may want to reconstruct the EnsembleKalmanProcess after every iteration, conserving the same prior. The current constructor did not allow this; it selected the current mean as the prior mean. This PR decouples the two, without changing the default behavior of the system.

In addition, comments and typing are improved for the module.

### DIFF
--- a/docs/src/parameter_distributions.md
+++ b/docs/src/parameter_distributions.md
@@ -45,7 +45,7 @@ x_eval = collect(-3:6/400:3)
 
 #bounded in [0.0, 1.0]
 transform_unconstrained_to_constrained(x) = exp(x) / (exp(x) + 1)
-dist= pdf(Normal(1, 0.5), x_eval)
+dist= pdf.(Normal(1, 0.5), x_eval)
 constrained_x_eval = transform_unconstrained_to_constrained.(x_eval)
 
 p1 = plot(x_eval, dist,) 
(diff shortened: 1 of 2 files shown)


Zhengyu-Huang (approved):
That looks great！ Thank you!

ilopezgp:
This is failing in the ParameterDistribution docs @navidcy , can you take a look since you modified that last?

navidcy:
I'll have a look @ilopezgp. I saw docs failing in other PRs, e.g. #74. I think there was a breaking change in one of the deps since #68 was building OK! I'll have a look at different package versions between then and now.

ilopezgp:
Thanks! I think it might be a change in the version of Distributions, this was the first error I saw:

`pdf(d::Distributions.Distribution{Distributions.ArrayLikeVariate{0},S} where S<:Distributions.ValueSupport, X::AbstractArray) in Distributions at deprecated.jl:70`

navidcy:
https://github.com/JuliaStats/Distributions.jl/issues/1447

navidcy:
I'm testing locally to build with Distributions v0.25.33. Will let you know!

navidcy:
It's not that. It's something else... beats me. I even tried with Distributions 0.25.19 (which was the version used in #68).

I'll try to go to the bottom of it. But this might be suggesting that we better add the Manifest.toml in the docs/ directory to have absolute control of all package versions.

ilopezgp:
@navidcy :) we just had to broadcast the method!

PR #81, opened 2021-12-05
Pass explicit RNG object to methods involving stochasticity

tsj5:
[*edited 19 Jan 2022*]

The motivation for this PR comes from CliMA/CalibrateEmulateSample.jl#121; see discussion there. 

In order to have reproducible tests involving randomness, it's not sufficient to set the global seed for the default RNG: the default RNG algorithm used by Julia was [changed](https://julialang.org/blog/2021/11/julia-1.7-highlights/#new_rng_reproducible_rng_in_tasks) in version 1.7, and the [docs](https://docs.julialang.org/en/v1/stdlib/Random/#Reproducibility) for the Random package note that the same seed may give different random values in different versions. For this reason, [recommended practice](https://discourse.julialang.org/t/set-the-state-of-the-global-rng/12599/3) for reproducible tests is to pass an RNG object explicitly to all methods which need it. Julia convention is that the RNG is an optional first positional argument in functions which require it (see e.g. Random, StatsBase, Turning.jl...), and this is done here.

This PR implements that change; in brief
- An `rng` field is added to `EnsembleKalmanProcesses` (i.e. independent of `Process`, for all implementations).
- An optional `rng` kwarg is added to `construct_initial_ensemble()`, used in lieu of `rng_seed` if provided.
- An additional method for `construct_initial_ensemble()` takes the rng as the first argument, causing kwargs to be ignored.
- All methods for `sample_distribution()` now accept an rng as an optional first argument.

In all of the cases above, `Random.GLOBAL_RNG` is used by default if an RNG is not given: this PR is fully backwards compatible with `main`. In particular, tests are updated to use an explicit rng, but examples have been left unchanged. It may be desirable to update them as well.


tsj5:
Updated description above. PR is ready to be merged if no one has further comments/questions.

odunbar:
Could you add unit tests to `test/ParameterDistribution/runtests.jl`for Line 307,308. should be 2 extra lines and will keep codecov happy. Then we can merge this

tsj5:
> Could you add unit tests to `test/ParameterDistribution/runtests.jl`for Line 307,308 in `src/ParameterDistributions`. 

Thanks, @odunbar, missed the codecov warning. 8056f72 adds tests for every method, so codecov should be happy now.

odunbar:
@tsj5 just the formatter

ilopezgp:
Let's rebase to test with Julia/1.7 and then squash all the commits into 1 for better tracking of the package. Thanks!

tsj5:
> Let's rebase to test with Julia/1.7 and then squash all the commits into 1 for better tracking of the package. Thanks!

Hi @ilopezgp, thanks for looking over this PR! Please be patient with me, as I'm still getting up to speed on CliMA's development conventions: I'm a bit scared of rewriting commit history, so I did the rebase and squash in a copy of this feature branch. I've submitted that as PR #95, which is intended to replace this PR (will close this one when that's accepted). If you'd rather have me overwrite this branch, let me know.

Thanks!

PR #86, opened 2021-12-10
Standardize UKI constructor to be consistent with EKS, document kwargs 

Zhengyu-Huang:
The UKI constructor modification

```
function Unscented(
    u0_mean::Vector{FT},
    uu0_cov::Matrix{FT};
    α_reg::FT = 1.0,
    update_freq::IT = 1,
    prior_mean::Union{Vector{FT}, Nothing} = nothing,
) where {FT <: AbstractFloat, IT <: Int}
```

require α_reg, update_freq, modified_unscented_transform, and  prior_mean as keyword arguments
clean up κ and β.


ilopezgp (approved):
LGTM

PR #91, opened 2022-01-18
Allow quadrature of scalar metrics in UKI, vectorize mean quadrature.

ilopezgp:
Adds the possibility to perform quadratures over scalars in UKI, and vectorizes some computations.

It also enables the use of `construct_cov()` directly without passing the mean. However, the default behavior is unchanged.


Zhengyu-Huang (approved):
The changes look correct! Thank you for the great work!

ilopezgp:
bors ping

PR #94, opened 2022-01-20
Update julia tests to 1.7

ilopezgp:


odunbar:
Does this need to wait for #81 to merge first?

ilopezgp:
Probably better to merge this one first. That way, #81 can be tested on both 1.6 and 1.7

Issue #55, opened 2021-09-22
Random seeds


odunbar:
We add noise during various updates, for example when the forward map is deterministic it is used to stabilize the method. Some users may require fixing random seeds, we have done this during the initial ensemble construction,  but not yet for the update of the ensemble. for example in EKI
e.g. https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/4252146c019a133bc3ef5daeeb7e912371be38f0/src/EnsembleKalmanInversion.jl#L77
It would be nice to have the random seeding option
Coauthored with @trontrytel

ilopezgp:
Closed by @tsj5 's PR #95 .

PR #95, opened 2022-01-22
Pass explicit RNG object to methods involving stochasticity

tsj5:
This is a rebase and squash of the feature branch for PR #81 and contains no new commits. The PR message for #81 is reproduced below.
<hr>
The motivation for this PR comes from CliMA/CalibrateEmulateSample.jl#121; see discussion there. 

In order to have reproducible tests involving randomness, it's not sufficient to set the global seed for the default RNG: the default RNG algorithm used by Julia was [changed](https://julialang.org/blog/2021/11/julia-1.7-highlights/#new_rng_reproducible_rng_in_tasks) in version 1.7, and the [docs](https://docs.julialang.org/en/v1/stdlib/Random/#Reproducibility) for the Random package note that the same seed may give different random values in different versions. For this reason, [recommended practice](https://discourse.julialang.org/t/set-the-state-of-the-global-rng/12599/3) for reproducible tests is to pass an RNG object explicitly to all methods which need it. Julia convention is that the RNG is an optional first positional argument in functions which require it (see e.g. Random, StatsBase, Turning.jl...), and this is done here.

This PR implements that change; in brief
- An `rng` field is added to `EnsembleKalmanProcesses` (i.e. independent of `Process`, for all implementations).
- An optional `rng` kwarg is added to `construct_initial_ensemble()`, used in lieu of `rng_seed` if provided.
- An additional method for `construct_initial_ensemble()` takes the rng as the first argument, causing kwargs to be ignored.
- All methods for `sample_distribution()` now accept an rng as an optional first argument.

In all of the cases above, `Random.GLOBAL_RNG` is used by default if an RNG is not given: this PR is fully backwards compatible with `main`. In particular, tests are updated to use an explicit rng, but examples have been left unchanged. It may be desirable to update them as well.

### DIFF
--- a/src/EnsembleKalmanProcess.jl
+++ b/src/EnsembleKalmanProcess.jl
@@ -46,6 +46,8 @@ struct EnsembleKalmanProcess{FT <: AbstractFloat, IT <: Int, P <: Process}
     Δt::Vector{FT}
     "the particular EK process (`Inversion` or `Sampler` or `Unscented`)"
     process::P
+    "Random number generator object (algorithm + seed) used for sampling and noise, for reproducibility. Defaults to `Random.GLOBAL_RNG`."
+    rng::AbstractRNG
 end
 
 # outer constructors
@@ -56,6 +58,7 @@ end
         obs_noise_cov::Array{FT, 2},
         process::P;
         Δt = FT(1),
+        rng::AbstractRNG = Random.GLOBAL_RNG
     ) where {FT <: AbstractFloat, P <: Process}
 
 Ensemble Kalman process constructor.
@@ -66,6 +69,7 @@ function EnsembleKalmanProcess(
     obs_noise_cov::Array{FT, 2},
     process::P;
     Δt = FT(1),
+    rng::AbstractRNG = Random.GLOBAL_RNG,
 ) where {FT <: AbstractFloat, P <: Process}
 
     #initial parameters stored as columns
@@ -80,7 +84,7 @@ function EnsembleKalmanProcess(
     # timestep store
     Δt = Array([Δt])
 
-    EnsembleKalmanProcess{FT, IT, P}([init_params], obs_mean, obs_noise_cov, N_ens, g, err, Δt, process)
+    EnsembleKalmanProcess{FT, IT, P}([init_params], obs_mean, obs_noise_cov, N_ens, g, err, Δt, process, rng)
 end
 
 
@@ -163,17 +167,24 @@ function get_N_iterations(ekp::EnsembleKalmanProcess)
 end
 
 """
-    construct_initial_ensemble(prior::ParameterDistribution, N_ens::IT; rng_seed=42) where {IT<:Int}
+    construct_initial_ensemble(prior::ParameterDistribution, N_ens::IT; rng_seed::IT = 42, rng::Union{Random.AbstractRNG,Nothing} = nothing) where {IT<:Int}
 
 Construct the initial parameters, by sampling `N_ens` samples from specified
 prior distribution. Returned with parameters as columns.
 """
-function construct_initial_ensemble(prior::ParameterDistribution, N_ens::IT; rng_seed = 42) where {IT <: Int}
-    # Ensuring reproducibility of the sampled parameter values
-    Random.seed!(rng_seed)
-    parameters = sample_distribution(prior, N_ens) #of size [dim(param space) N_ens]
-    return parameters
+function construct_initial_ensemble(
+    prior::ParameterDistribution,
+    N_ens::IT;
+    rng_seed::IT = 42,
+    rng::Union{AbstractRNG, Nothing} = nothing,
+) where {IT <: Int}
+    # Ensuring reproducibility of the sampled parameter values: re-seed GLOBAL_RNG if we're
+    # given a seed, but if we got an explicit rng, we shouldn't re-seed it
+    rng = isnothing(rng) ? Random.seed!(rng_seed) : rng
+    return sample_distribution(rng, prior, N_ens) #of size [dim(param space) N_ens]
 end
+construct_initial_ensemble(rng::AbstractRNG, prior::ParameterDistribution, N_ens::IT) where {IT <: Int} =
+    construct_initial_ensemble(prior, N_ens; rng = rng)
 
 function compute_error!(ekp::EnsembleKalmanProcess)
     mean_g = dropdims(mean(get_g_final(ekp), dims = 2), dims = 2)
--- a/test/EnsembleKalmanProcess/runtests.jl
+++ b/test/EnsembleKalmanProcess/runtests.jl
@@ -12,7 +12,7 @@ const EKP = EnsembleKalmanProcesses
 
     # Seed for pseudo-random number generator
     rng_seed = 42
-    Random.seed!(rng_seed)
+    rng = Random.MersenneTwister(rng_seed)
 
     ### Generate data from a linear model: a regression problem with n_par parameters
     ### and 1 observation of G(u) = A \times u, where A : R^n_par -> R^n_obs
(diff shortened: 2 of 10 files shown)


src/EnsembleKalmanProcess.jl
```diff
@@ -163,17 +167,26 @@ function get_N_iterations(ekp::EnsembleKalmanProcess)
...
+    rng_seed::IT = 42,
+    rng::Union{Random.AbstractRNG, Nothing} = nothing,
+) where {IT <: Int}
     # Ensuring reproducibility of the sampled parameter values
-    Random.seed!(rng_seed)
-    parameters = sample_distribution(prior, N_ens) #of size [dim(param space) N_ens]
-    return parameters
+    if rng === nothing
+        rng = Random.seed!(rng_seed)
+    end
```
ilopezgp, line 184:
```suggestion
    rng = isnothing(rng) ? Random.seed!(rng_seed) : rng
```

src/EnsembleKalmanProcess.jl
```diff
@@ -163,17 +167,26 @@ function get_N_iterations(ekp::EnsembleKalmanProcess)
...
+    # on the other hand, if we did pass an explicit rng, we seeded it already
```
ilopezgp, line 185: move comment before `rng` check.

src/EnsembleKalmanProcess.jl
```diff
@@ -163,17 +167,26 @@ function get_N_iterations(ekp::EnsembleKalmanProcess)
...
+    return sample_distribution(rng, prior, N_ens) #of size [dim(param space) N_ens]
 end
+construct_initial_ensemble(rng::AbstractRNG, prior::ParameterDistribution, N_ens::IT) where {IT <: Int} =
```
ilopezgp, line 188: Sometimes you use `AbstractRNG`, and others `Random.AbstractRNG`. Can you homogenize to `AbstractRNG`?

test/EnsembleKalmanProcessModule/runtests.jl
```diff
@@ -0,0 +1,219 @@
+using Distributions
+using LinearAlgebra
+using Random
+using Test
+
+using EnsembleKalmanProcesses.EnsembleKalmanProcessModule
+using EnsembleKalmanProcesses.ParameterDistributionStorage
+@testset "EnsembleKalmanProcessModule" begin
```
ilopezgp, line 8: These need to be updated to the new module notation.

ilopezgp (commented):
Great addition @tsj5 !!

test/EnsembleKalmanProcessModule/runtests.jl
tsj5: Actually, missed that `test/EnsembleKalmanProcessModule/runtests.jl` was renamed to `test/EnsembleKalmanProcess/runtests.jl` in `main` (and edited further). In c1b164d I implemented the changes I had made to the EKPModule tests in the EKP tests, and deleted the EKPModule tests. Verified that updated tests pass on 1.6 and 1.7.

tsj5:
> Great addition @tsj5 !!

Thanks for the detailed review! All changes suggested above have been made in c1b164d (assuming you wanted to keep things squished).

src/EnsembleKalmanProcess.jl
tsj5: Done in c1b164d

src/EnsembleKalmanProcess.jl
tsj5: Done in c1b164d

src/EnsembleKalmanProcess.jl
tsj5: Done in c1b164d

test/EnsembleKalmanProcess/runtests.jl
```diff
@@ -158,7 +165,7 @@ const EKP = EnsembleKalmanProcesses
         eki_final_result = vec(mean(get_u_final(ekiobj), dims = 2))
 
         # kind of arbitrary tolerance here,
-        @test norm(u_star - eki_final_result) < 0.1
+        @test norm(u_star - eki_final_result) < 0.15
```
tsj5, line 168: This test is the reason CI [failed on Ubuntu](https://github.com/CliMA/EnsembleKalmanProcesses.jl/runs/4909239465?check_suite_focus=true#step:6:172) in previous commit; missed this because I don't have an Ubuntu box to test on.

Based on the comment, I'm increasing the tolerance so that the test passes. For the record, in the earlier version of these tests (EnsembleKalmanProcessModule/runtests.jl) this test was commented out, and the tolerance was set to 0.5.

ilopezgp (approved):
LGTM

test/EnsembleKalmanProcess/runtests.jl
ilopezgp: Mmmmh this seems to still be a problem sometimes. Honestly, this was a bad test to begin with. Could you instead compare
`@test norm(u_star - eki_final_result) < norm(u_star - eki_init_result)`?

where
eki_init_result = vec(mean(get_u_prior(ekiobj), dims = 2))

tsj5: Thanks, that's a better idea than just bumping the threshold!

I'll continue squashing commits onto this branch and copying/cherry-picking them onto `tsj/add-explicit-rng` #81, to have a paper trail -- let me know if you want me to proceed differently. Apologies for making this messier than it needed to be: I understand that you want to have the PR reviewed and finalized *before* the squash... 🙁

tsj5: Done in b55846d

tsj5:
I'm a bit confused by [CI results](https://github.com/CliMA/EnsembleKalmanProcesses.jl/runs/4909316258?check_suite_focus=true) on the previous commit: the [problematic EKP unit test](https://github.com/tsj5/EnsembleKalmanProcesses.jl/blob/e7dbdc44bdfd5f4f6efa12554d1c4dbf6a9ddd2e/test/EnsembleKalmanProcess/runtests.jl#L168) passes for julia-1.7/ubuntu but fails for julia-latest/ubuntu, even though those test envs use the *same* version of julia and ubuntu. This is troubling, since the entire purpose of this PR is to make the package more reproducible.

I found some places where the rng wasn't being passed through, and fixed them in e4bb0d467c1b10900f39c3e07ad3b16ff29112d7 (squashed here in b55846d). The changes aren't in the code path for the failing test, though. I'll keep looking into this non-reproducibility as best as I can, since I don't have an ubuntu image handy.

ilopezgp:
> I'm a bit confused by [CI results](https://github.com/CliMA/EnsembleKalmanProcesses.jl/runs/4909316258?check_suite_focus=true) on the previous commit: the [problematic EKP unit test](https://github.com/tsj5/EnsembleKalmanProcesses.jl/blob/e7dbdc44bdfd5f4f6efa12554d1c4dbf6a9ddd2e/test/EnsembleKalmanProcess/runtests.jl#L168) passes for julia-1.7/ubuntu but fails for julia-latest/ubuntu, even though those test envs use the _same_ version of julia and ubuntu. This is troubling, since the entire purpose of this PR is to make the package more reproducible.
> 
> I found some places where the rng wasn't being passed through, and fixed them in [e4bb0d4](https://github.com/CliMA/EnsembleKalmanProcesses.jl/commit/e4bb0d467c1b10900f39c3e07ad3b16ff29112d7) (squashed here in [b55846d](https://github.com/CliMA/EnsembleKalmanProcesses.jl/commit/b55846d2801b73edac81e46faad66c5d20d33ada)). The changes aren't in the code path for the failing test, though. I'll keep looking into this non-reproducibility as best as I can, since I don't have an ubuntu image handy.

That's a good point! It seems one of the random seems is changing between architectures, Julia versions, or maybe just instances of the test.

ilopezgp:
Closes #55

Issue #92, opened 2022-01-19
Simplify interface


charleskawczynski:
One of the nice parts of Julia is that we can extend existing methods, for example, 

```julia
get_mean(d::Parameterized) = mean(d.distribution)
```
could simply extend the `StatsBase` `mean` method
```julia
import StatsBase
...
StatsBase.mean(d::Parameterized) = StatsBase.mean(d.distribution)
```
This can be applied to several methods (e.g., `get_mean`, `len`, `size` etc.)

ilopezgp:
Let's go through these before release `0.2`, since some of these will be breaking changes.

PR #96, opened 2022-01-25
Bump patch release

navidcy:


ilopezgp:
This is a breaking release, it cannot be a patch. We are waiting on #92 to be resolved so that we limit breaking to the next release for a while.

PR #98, opened 2022-01-25
Extending functions from packges

### LINKED ISSUE #92: Simplify interface

odunbar:
## Purpose
To replace some functions with corresponding extensions from existing packages. (and the relevant tests/examples)
resolves #92 
## In this PR
- ParameterDistributions: By extending from `StatsBase`. Replacing `get_mean` `get_var` `get_cov` 'sample_distribution' by extending `mean` `var` `cov` `sample` 
- ParameterDistributions: by extending `Base`: Change `len` to `length`, added `size`, change `dimension`and `get_total_dimension` to `ndims`

### DIFF
--- a/docs/src/API/ParameterDistributions.md
+++ b/docs/src/API/ParameterDistributions.md
@@ -23,8 +23,9 @@ batch
 get_distribution
 sample_distribution
 get_logpdf
-get_cov
-get_mean
+mean
+var
+cov
 transform_constrained_to_unconstrained
 transform_unconstrained_to_constrained
 ```
--- a/docs/src/API/ParameterDistributions.md
+++ b/docs/src/API/ParameterDistributions.md
@@ -22,7 +22,7 @@ get_n_samples
 get_all_constraints
 batch
 get_distribution
-sample_distribution
+sample
 get_logpdf
 mean
 var
--- a/docs/src/API/ParameterDistributions.md
+++ b/docs/src/API/ParameterDistributions.md
@@ -8,13 +8,14 @@ CurrentModule = EnsembleKalmanProcesses.ParameterDistributions
 Parameterized
 Samples
 Constraint
+ParameterDistribution
 no_constraint
 bounded_below
 bounded_above
 bounded
-len
+length(c::CType) where {CType <: ConstraintType}
+size(c::CType) where {CType <: ConstraintType}
 n_samples
-ParameterDistribution
 get_name
 get_dimensions
 get_n_samples
--- a/src/ParameterDistributions.jl
+++ b/src/ParameterDistributions.jl
@@ -7,7 +7,7 @@ using Random
 
 #import (to add definitions)
 import StatsBase: mean, var, cov, sample
-import Base: size, length
+import Base: size, length, ndims
 ## Exports
 
 #types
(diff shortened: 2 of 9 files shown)


ilopezgp:
I saw a few instances of the pythonic `len` instead of the julian `length`. Let's change those as well.

odunbar:
bors try

docs/src/API/ParameterDistributions.md
```diff
@@ -8,13 +8,14 @@ CurrentModule = EnsembleKalmanProcesses.ParameterDistributions
...
 Samples
 Constraint
+ParameterDistribution
 no_constraint
 bounded_below
 bounded_above
 bounded
-len
+length(c::CType) where {CType <: ConstraintType}
+size(c::CType) where {CType <: ConstraintType}
```
ilopezgp, line 17: Is this required because the methods are from `Base`? I am trying to understand the difference between these and `mean`, `var`, etc, which expand upon StatsBase.

src/ParameterDistributions.jl
```diff
@@ -21,8 +23,8 @@ export get_name, get_distribution, get_total_dimension, get_dimensions, get_all_
 export sample_distribution
 export no_constraint, bounded_below, bounded_above, bounded
 export transform_constrained_to_unconstrained, transform_unconstrained_to_constrained
-export get_logpdf, get_cov, get_var, get_mean, batch
-
+export get_logpdf, batch
+#export get_cov, get_var, get_mean
```
ilopezgp, line 27: can remove this.

src/ParameterDistributions.jl
```diff
@@ -282,7 +291,7 @@ optional and defaults to 1.
 """
 function sample_distribution(rng::AbstractRNG, d::Samples, n_draws::IT) where {IT <: Integer}
```
ilopezgp, line 292: Could sample_distribution extend `sample`?

docs/src/API/ParameterDistributions.md
odunbar: yeah docs gave an error that there was duplication in the `length` and `size` when i didn't have the explicit method.

src/ParameterDistributions.jl
odunbar: thanks - good suggestion

docs/src/API/ParameterDistributions.md
odunbar: I wonder if it doesnt give an error with the clash in `StatsBase` because this package is not a dependence in docs `Project.toml` (if this is what you are surprised about)

src/ParameterDistributions.jl
```diff
@@ -118,33 +120,39 @@ function bounded(lower_bound::FT, upper_bound::FT) where {FT <: Real}
...
-    get_dimensions(d<:ParametrizedDistributionType)
+    ndims(d<:ParametrizedDistributionType)
 
 The number of dimensions of the parameter space
 """
-dimension(d::Parameterized) = length(d.distribution)
+ndims(d::Parameterized) = length(d.distribution)
 
-dimension(d::Samples) = size(d.distribution_samples)[1]
+ndims(d::Samples) = size(d.distribution_samples)[1]
```
ilopezgp, line 146:
```suggestion
ndims(d::Samples) = size(d.distribution_samples, 1)
```

ilopezgp (approved):
Thank you @odunbar !

Issue #99, opened 2022-01-26
Covariance matrix `obs_noise_cov` is restricted to `Array{FT, 2}`


glwagner:
The type of `EnsembleKalmanProcess.obs_noise_cov` is restricted to `Array{FT, 2}`, which is inconvenient and prevents the use of common `LinearAlgebra` abstractions that represent arrays.

For example:

```julia
julia> using EnsembleKalmanProcesses, EnsembleKalmanProcesses.ParameterDistributions, LinearAlgebra, Distributions

julia> distribution = ParameterDistribution(Parameterized(Normal(0, 1)), no_constraint(), "");

julia> sample = sample_distribution(distribution, 1);

julia> obs_noise_cov = 0.1*I
UniformScaling{Float64}
0.1*I

julia> ekp = EnsembleKalmanProcess(sample, zeros(2), obs_noise_cov, Inversion())
ERROR: MethodError: no method matching EnsembleKalmanProcess(::Matrix{Float64}, ::Vector{Float64}, ::UniformScaling{Float64}, ::Inversion)
...
```

`0.1 * I` is a valid and efficient way to specify a covariance matrix. The code itself likely works as is, since key lines like

```julia
julia> noise = rand(MvNormal(zeros(2), 0.1 * I), 2)
2×2 Matrix{Float64}:
 0.0551659  0.720873
 0.057236   0.289555
```

are valid for a wide variety of matrix types. Thus probably the only change needed is to release `obs_noise_cov` from `Array{FT, 2}`.

For large observation vectors the need to convert objects like `UniformScaling` to `Array{FT, 2}` is a waste of memory:

```julia
julia> Matrix(0.1*I, 5, 5)
5×5 Matrix{Float64}:
 0.1  0.0  0.0  0.0  0.0
 0.0  0.1  0.0  0.0  0.0
 0.0  0.0  0.1  0.0  0.0
 0.0  0.0  0.0  0.1  0.0
 0.0  0.0  0.0  0.0  0.1
```

Another important matrix type is `Diagonal` representing a diagonal matrix.

cc @simonbyrne

odunbar:
For sure, the creation of the package was done with types being used to restrict arguments (before i even knew about the difficulties with matrix != array{,2} != DiagonalMatrix type problems) 

I think for Covariance matrices, I am happy to take the broadest umbrella type that can have a dimension specified (i.e. must be 2D)?

glwagner:
[...]

PR #100, opened 2022-01-26
Substitute types by abstractions.

ilopezgp:
Relaxes the allowed types of some objects and input args for improved generalization:

- Vector - > AbstractVector
- Matrix -> AbstractMatrix
- Sparse.SparseMatrix -> AbstractMatrix
- Union{Vector, Matrix} -> AbstractVecOrMat

For covariance matrices in output space,

- Matrix -> Union{AbstractMatrix, UniformScaling}

Tests involving different matrix types are now included for all Ensemble Kalman Processes.

### DIFF
--- a/src/DataContainers.jl
+++ b/src/DataContainers.jl
@@ -16,9 +16,9 @@ Container to store data samples as columns in an array.
 """
 struct DataContainer{FT <: Real}
     #stored data, each piece of data is a column [data dimension × number samples]
-    stored_data::Array{FT, 2}
+    stored_data::AbstractMatrix{FT}
     #constructor with 2D arrays
-    function DataContainer(stored_data::Array{FT, 2}; data_are_columns = true) where {FT <: Real}
+    function DataContainer(stored_data::AbstractMatrix{FT}; data_are_columns = true) where {FT <: Real}
 
         if data_are_columns
             new{FT}(stored_data)
@@ -41,8 +41,8 @@ struct PairedDataContainer{FT <: Real}
 
     #constructor with 2D Arrays
     function PairedDataContainer(
-        inputs::Array{FT, 2},
-        outputs::Array{FT, 2};
+        inputs::AbstractMatrix{FT},
+        outputs::AbstractMatrix{FT};
         data_are_columns = true,
     ) where {FT <: Real}
 
--- a/src/ParameterDistributions.jl
+++ b/src/ParameterDistributions.jl
@@ -44,11 +44,11 @@ end
 A distribution comprised of only samples, stored as columns of parameters.
 """
 struct Samples{FT <: Real} <: ParameterDistributionType
-    distribution_samples::Array{FT, 2} #parameters are columns
-    Samples(distribution_samples::Array{FT, 2}; params_are_columns = true) where {FT <: Real} =
+    distribution_samples::AbstractMatrix{FT} #parameters are columns
+    Samples(distribution_samples::AbstractMatrix{FT}; params_are_columns = true) where {FT <: Real} =
         params_are_columns ? new{FT}(distribution_samples) : new{FT}(permutedims(distribution_samples, (2, 1)))
     #Distinguish 1 sample of an ND parameter or N samples of 1D parameter, and store as 2D array  
-    Samples(distribution_samples::Array{FT, 1}; params_are_columns = true) where {FT <: Real} =
+    Samples(distribution_samples::AbstractVector{FT}; params_are_columns = true) where {FT <: Real} =
         params_are_columns ? new{FT}(reshape(distribution_samples, 1, :)) : new{FT}(reshape(distribution_samples, :, 1))
 end
 
@@ -328,9 +328,9 @@ sample(d::Parameterized) = sample(Random.GLOBAL_RNG, d, 1)
 Obtains the independent logpdfs of the parameter distributions at `xarray`
 (non-Samples Distributions only), and returns their sum.
 """
-get_logpdf(d::Parameterized, xarray::Array{FT, 1}) where {FT <: Real} = logpdf.(d.distribution, xarray)
+get_logpdf(d::Parameterized, xarray::AbstractVector{FT}) where {FT <: Real} = logpdf.(d.distribution, xarray)
 
-function get_logpdf(pd::ParameterDistribution, xarray::Array{FT, 1}) where {FT <: Real}
+function get_logpdf(pd::ParameterDistribution, xarray::AbstractVector{FT}) where {FT <: Real}
     #first check we don't have sampled distribution
     for d in pd.distributions
         if typeof(d) <: Samples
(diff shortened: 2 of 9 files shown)


src/DataContainers.jl
```diff
@@ -16,9 +16,9 @@ Container to store data samples as columns in an array.
 """
 struct DataContainer{FT <: Real}
     #stored data, each piece of data is a column [data dimension × number samples]
-    stored_data::Array{FT, 2}
+    stored_data::AbstractMatrix{FT}
```
charleskawczynski, line 19: https://docs.julialang.org/en/v1/manual/performance-tips/#Avoid-fields-with-abstract-containers

jakebolewski: I would say most of infa here isn't performance critical

ilopezgp: A concern that can be limiting for the package is unnecessary memory use. In some cases, we want to store covariances that are extremely sparse, or even diagonal. Casting them and storing them as a Matrix{FT} then takes up much more space than e.g. Diagonal {FT}. Whichever solution solves this problem would work. Let me know how to proceed.

jakebolewski: I would say write it as generically as possible and then if there is some performance hotspots after a bit of profiling they should be straightforward to address.  You don't need to specialize on every Matrix type here so IMO what is written is fine.

src/ParameterDistributions.jl
```diff
@@ -452,7 +464,7 @@ Here, `xarray` contains parameters sample ensembles for different EKP iterations
 """
 function transform_unconstrained_to_constrained(
     pd::ParameterDistribution,
-    xarray::Array{Array{FT, 2}, 1},
+    xarray::AbstractVector{Matrix{FT}},
```
jakebolewski, line 467: I would leave this untyped as you just need to support an iterable that returns (abstract) matrices

src/ParameterDistributions.jl
```diff
@@ -440,7 +449,10 @@ end
 Apply the transformation to map parameter samples `xarray` from the unconstrained space into (possibly constrained) space.
 Here, `xarray` contains parameters as columns and samples as rows.
 """
-function transform_unconstrained_to_constrained(pd::ParameterDistribution, xarray::Array{FT, 2}) where {FT <: Real}
+function transform_unconstrained_to_constrained(
+    pd::ParameterDistribution,
+    xarray::AbstractMatrix{FT},
```
jakebolewski, line 454: I would leave this untyped as well

src/UnscentedKalmanInversion.jl
```diff
@@ -7,23 +7,23 @@ An unscented Kalman Inversion process.
 """
 mutable struct Unscented{FT <: AbstractFloat, IT <: Int} <: Process
     "a vector of arrays of size `N_parameters` containing the mean of the parameters (in each `uki` iteration a new array of mean is added)"
-    u_mean::Vector{Vector{FT}}
+    u_mean::AbstractVector{Vector{FT}}
```
jakebolewski, line 10: I would leave this field untyped as you just need an iterable of some vector type

src/UnscentedKalmanInversion.jl
```diff
@@ -7,23 +7,23 @@ An unscented Kalman Inversion process.
...
     "a vector of arrays of size (`N_parameters x N_parameters`) containing the covariance of the parameters (in each `uki` iteration a new array of `cov` is added)"
-    uu_cov::Vector{Matrix{FT}}
+    uu_cov::AbstractVector{Matrix{FT}}
```
jakebolewski, line 12: Same here (for all container of container types)

src/UnscentedKalmanInversion.jl
```diff
@@ -7,23 +7,23 @@ An unscented Kalman Inversion process.
...
     "a vector of arrays of size `N_y` containing the predicted observation (in each `uki` iteration a new array of predicted observation is added)"
-    obs_pred::Vector{Vector{FT}}
+    obs_pred::AbstractVector{Vector{FT}}
```
jakebolewski, line 14: and here

src/DataContainers.jl
charleskawczynski: I wasn't suggesting to specialize on every type of matrix, just to make the type (and its properties) concrete. This was @glwagner's [suggestion](https://github.com/CliMA/EnsembleKalmanProcesses.jl/issues/99#issuecomment-1022773454), too. It's not much overhead to have very concrete types and simply not specialize on methods that use it.

jakebolewski: then you cannot mix and mach representations (different matrix types might be required for different fields).  Until performance is actually an issue I wouldn't worry about it

jakebolewski: for calibration the costs are going to be dominated by actually running the models and IO, everything else I'm guessing won't matter much but we'll see as we go

src/ParameterDistributions.jl
ilopezgp: @jakebolewski I can leave the one above untyped, but I think I need to specify the type on this one because we are using the type for dispatch. We have three `transform_unconstrained_to_constrained`, and I presume only one of them can be the default untyped version.

jakebolewski: ah ok, you are correct (didn't see that from the diff)

src/DataContainers.jl
glwagner: > A concern that can be limiting for the package is unnecessary memory use. In some cases, we want to store covariances that are extremely sparse, or even diagonal. Casting them and storing them as a Matrix{FT} then takes up much more space than e.g. Diagonal {FT}. Whichever solution solves this problem would work. Let me know how to proceed.

@ilopezgp the point that @charleskawczynski was making is the difference between

```julia
struct DataContainer{FT <: Real}
    #stored data, each piece of data is a column [data dimension × number samples]
    stored_data::Array{FT, 2}
    stored_data::AbstractMatrix{FT}
```

and 

```julia
struct DataContainer{M <: AbstractMatrix}
    #stored data, each piece of data is a column [data dimension × number samples]
    stored_data::M
```

In the first case, `stored_data` is _abstractly typed_, which means that the compiler cannot infer the type of `stored_data` given the type `DataContainer`. This has performance penalties (which @jakebolewski argues are not important). But it's utterly fundamental and so an important point to grasp.

Writing `M <: AbstractMatrix` will preclude `UniformScaling`. Omitting `<: AbstractMatrix` is the conservative choice. You might also write `M <: Union{AbstractMatrix, UniformScaling}`. However this is risky because you may be failing to anticipate other valid matrix-like types.

ilopezgp: I understand the difference between abstract types and parameterized types, thank you. I think the conversation was about whether it practically matters in this application.

odunbar:
Could I ask that you add the specific changes to the PR as a reference for the conventions we take? i see currently
- Vector - > AbstractVector
- Matrix -> AbstractMatrix
- Sparse.SparseMatrix -> AbstractMatrix
- Union{Vector, Matrix} -> AbstractVecOrMat
- Vector{Vector} -> Iterable{Vector}
And on from Greg's point, we currently will exclude UniformScaling. (is this just for aesthetic?)

ilopezgp:
> Could I ask that you add the specific changes to the PR comment as a reference for the conventions we take? i see currently
> 
> * Vector - > AbstractVector
> * Matrix -> AbstractMatrix
> * Sparse.SparseMatrix -> AbstractMatrix
> * Union{Vector, Matrix} -> AbstractVecOrMat
> * Vector{Vector} -> Iterable{Vector}
> 
> And on from Greg's point, we currently will exclude UniformScaling. (is this just for aesthetic?)

UniformScaling was in fact a specific request in issue #99, so we should include it.

odunbar (approved):
This is a good step towards greater flexibility! LGTM and addresses #99

ilopezgp:
Closes #99 (the specific problems initially raised).

PR #102, opened 2022-01-27
Update Tests.yml

jakebolewski:
reduce the number of CI runs


ilopezgp (approved):
LGTM. We changed it to make sure we were testing Julia 1.7, which is the compat of other packages, but I guess we will be updating all other packages to the latest Julia once it stops being 1.7 as well.

Issue #104, opened 2022-02-04
Deterministic behavior of stochastic functions should not be the default


ilopezgp:
A consequence of #95 is that now, some functions that should be stochastic by default are deterministic.

An example of this is the following,

```
using Distributions
using Random

using EnsembleKalmanProcesses
using EnsembleKalmanProcesses.ParameterDistributions

d = Parameterized(Normal(0, 1))
c = no_constraint()
name = "unconstrained_normal"
u = ParameterDistribution(d, c, name)

for i in 1:5
   println(construct_initial_ensemble(u, 1))
end
```

This outputs,

```
[0.7883556016042917;;]
[0.7883556016042917;;]
[0.7883556016042917;;]
[0.7883556016042917;;]
[0.7883556016042917;;]
```

which means that we are no longer sampling i.i.d from the distribution. This is desirable for reproducibility of tests, but **not** as the default behavior. The expected user behavior can be achieved at the moment using

```
for i in 1:5
       println(construct_initial_ensemble(u, 1, rng= Random.GLOBAL_RNG))
end
```

which for example can output

```
[0.9548452855545422;;]
[0.3785684038407532;;]
[-0.257360852691082;;]
[-1.1705533137462902;;]
[0.9307323496630826;;]
```

Proposed solution: Make Random.GLOBAL_RNG the default AbstractRNG kwarg for all functions in the package, pass rng when the user wants reproducibility (e.g., test suite)

ilopezgp:
@tsj5 let me know if you have bandwidth to solve this issue

tsj5:
@ilopezgp oof, that's a bad bug, didn't mean to break the rng entirely! Thanks for catching this -- was caused by not thinking through the consequences of default kwargs, as you pointed out. Will have a PR to fix it in a few min...

PR #105, opened 2022-02-05
Correct rng kwargs for construct_initial_ensemble()

tsj5:
Fixes CliMA/EnsembleKalmanProcesses.jl#104.

 This PR corrects the logic for construct_initial_ensemble() to the following:
1. If an rng is not passed as the first argument, GLOBAL_RNG is used (same convention as StatsBase, Random, etc.)
2. If and only if a value for `rng_seed` is given, the rng is re-seeded to that value.

Previously, passing both `rng_seed` and `rng` as optional kwargs to  meant that, if neither was supplied, GLOBAL_RNG was re-seeded to 42 on every call.

Add @ilopezgp's MWE from #104 to the unit tests as a sanity check.

### DIFF
--- a/src/EnsembleKalmanProcess.jl
+++ b/src/EnsembleKalmanProcess.jl
@@ -167,24 +167,27 @@ function get_N_iterations(ekp::EnsembleKalmanProcess)
 end
 
 """
-    construct_initial_ensemble(prior::ParameterDistribution, N_ens::IT; rng_seed::IT = 42, rng::Union{Random.AbstractRNG,Nothing} = nothing) where {IT<:Int}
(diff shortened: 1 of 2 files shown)


ilopezgp (approved):
Thank you @tsj5 !! Was this the only place where the random seed was fixed by construction?

tsj5:
> Thank you @tsj5 !! Was this the only place where the random seed was fixed by construction in the default?

Yes -- did a full-text search for "seed" in my IDE. The only methods in EKP using an rng are construct_initial_ensemble() and sample() [renamed from sample_distribution(), in pre-0.2].

ilopezgp:
Closes #104

PR #109, opened 2022-02-08
Add continuous UKI dynamics discretization.

ilopezgp:
Adds the continuous version of UKI dynamics as specified in Huang et al (2021), section 3.4,

Σ_ω -> Σ_ω * Δt
Σ_ν -> Σ_ν / Δt

I didn't change alpha to alpha_h, I think that's fine from the paper.

 Also moves the dt constructor to EnsembleKalmanProcess.jl.


Zhengyu-Huang (approved):
I think the changes are correct! Thank you for the great work!

PR #112, opened 2022-02-25
Add failsafe EKI with sample_succ_gauss handler.

ilopezgp:
- Adds a failure handler struct to all `EnsembleKalmanProcess` objects.
- Implements the `IgnoreFailures` failure handler for all `EnsembleKalmanProcess`es, which maintains the former behavior of the algorithms (i.e, do nothing upon failure).
- Implements the `SampleSuccGauss` failure handler for the Inversion process, which advances the failed ensemble by sampling from an empirical gaussian formed by the successful ensemble.
- Failure may be defined by the user by passing the indices of failed particles to `update_ensemble!`. If nothing is passed, failures are taken to be those forward model evaluations that contain NaNs.
- Adds tests of new failure handlers.

In addition:

- Update equations are encapsulated outside of `update_ensemble!`, and into functions `..._update`.
- Transposes in the Sampler are encapsulated to the failsafe_update call. This allows to work with data as columns in the Sampler's version of `update_ensemble!`.

### DIFF
--- a/src/EnsembleKalmanInversion.jl
+++ b/src/EnsembleKalmanInversion.jl
@@ -7,6 +7,32 @@ An ensemble Kalman Inversion process
 """
 struct Inversion <: Process end
 
+function FailureHandler(process::Inversion, method::IgnoreFailures)
+    failsafe_update(u, g, y, obs_noise_cov, failed_ens) = eki_update(u, g, y, obs_noise_cov)
+    return FailureHandler{Inversion, IgnoreFailures}(failsafe_update)
+end
+
+"""
+    FailureHandler(process::Inversion, method::SampleSuccGauss)
+
+Provides a failsafe update that
+ - updates the successful ensemble according to the EKI update,
+ - updates the failed ensemble by sampling from the updated successful ensemble.
+"""
+function FailureHandler(process::Inversion, method::SampleSuccGauss)
+    function failsafe_update(u, g, y, obs_noise_cov, failed_ens)
+        successful_ens = filter(x -> !(x in failed_ens), collect(1:size(g, 2)))
+        n_failed = length(failed_ens)
+        u[:, successful_ens] =
+            eki_update(u[:, successful_ens], g[:, successful_ens], y[:, successful_ens], obs_noise_cov)
+        if !isempty(failed_ens)
+            u[:, failed_ens] = sample_empirical_gaussian(u[:, successful_ens], n_failed)
+        end
+        return u
+    end
+    return FailureHandler{Inversion, SampleSuccGauss}(failsafe_update)
+end
+
 """
     find_ekp_stepsize(ekp::EnsembleKalmanProcess{FT, IT, Inversion}, g::AbstractMatrix{FT}; cov_threshold::FT=0.01) where {FT}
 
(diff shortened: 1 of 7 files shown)


src/EnsembleKalmanInversion.jl
```diff
@@ -63,13 +144,10 @@ function update_ensemble!(
 
     # u: N_par × N_ens 
     # g: N_obs × N_ens
-    u = get_u_final(ekp)
+    u = deepcopy(get_u_final(ekp))
```
odunbar, line 147: https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/8c3be1d253cef55df27ec1e88905d6a8d226f76a/src/DataContainers.jl#L96

All get methods with a `DataContainer` object call a deepcopy already so this is not required

src/EnsembleKalmanInversion.jl
```diff
@@ -80,10 +158,28 @@ function update_ensemble!(
     # G is deterministic
     y = deterministic_forward_map ? (ekp.obs_mean .+ noise) : (ekp.obs_mean .+ zero(noise))
 
-    # N_obs × N_obs \ [N_obs × N_ens]
-    # --> tmp is [N_obs × N_ens]
-    tmp = (cov_gg + scaled_obs_noise_cov) \ (y - g)
-    u += (cov_ug * tmp) # [N_par × N_ens]
+    if isnothing(failure_handler)
```
odunbar, line 161: Rather than the if-else's and strings. It might be nice to wrap up error handling a little more, I was wondering about floating the idea of this:
```julia
struct FailureHandler{Process::P}
    update_failure_method::Function
end
function GaussianFailure(Process::P)
    update_failure_method = ...
    return FailureHandler{P}(update_failure_method)
end
```
then have in the update we call  `handle_failures(u,g,y,cov,failure_handler)` that does something like this
```julia 
function handle_failures(u,g,y,cov,failure_handler)
    failure_handler.update_failure_method(u,g,y,cov)
end
```
Or something else, I'm not quite sure yet what would be smoothest across the different methods but this sort of thing could be nice! Any thoughts?

The benefit here would be that someone can add new failure methods without tampering with the updates, or if they want parameters in the failure methods, this will not change the input arguments to the update_ensemble.

ilopezgp:
This works for me locally, any idea what the error means? @charleskawczynski @jakebolewski

src/EnsembleKalmanProcess.jl
```diff
@@ -209,6 +210,49 @@ function set_Δt!(ekp::EnsembleKalmanProcess, Δt_new::T) where {T}
...
+struct FailureHandler{P <: Process, FM <: FailureHandlingMethod}
+    failsafe_update::Function
+end
+
+"""
+     sample_empirical_gaussian(u::AbstractMatrix{FT}, n::IT) where {FT <: Real, IT}
+
+Returns `n` samples from an empirical Gaussian based on point estimates `u`.
+"""
+function sample_empirical_gaussian(u::AbstractMatrix{FT}, n::IT; inflation::FT = eps(Float64)) where {FT <: Real, IT}
```
charleskawczynski, line 227:
```suggestion
function sample_empirical_gaussian(u::AbstractMatrix{FT}, n::IT; inflation::FT = eps(FT)) where {FT <: Real, IT}
```

src/EnsembleKalmanInversion.jl
```diff
@@ -80,10 +141,15 @@ function update_ensemble!(
...
 
-    # N_obs × N_obs \ [N_obs × N_ens]
-    # --> tmp is [N_obs × N_ens]
-    tmp = (cov_gg + scaled_obs_noise_cov) \ (y - g)
-    u += (cov_ug * tmp) # [N_par × N_ens]
+    if isnothing(failed_ens)
+        _, failed_ens = split_indices_by_success(g)
+    end
+    if !isempty(failed_ens)
+        @info "$(length(failed_ens)) particle failure(s) detected. Handler used: $failure_handler_method."
```
charleskawczynski, line 148:
```suggestion
        @info "$(length(failed_ens)) particle failure(s) detected. Handler used: $(nameof(FM))."
```
This may be more human-readable

charleskawczynski:
I'm getting a seg fault, I can try taking a closer look tomorrow

ilopezgp:
> I'm getting a seg fault, I can try taking a closer look tomorrow

Thanks! That would be helpful. Are you getting seg fault locally? All the tests run for me locally (MacOS), but fail here

charleskawczynski:
> Are you getting seg fault locally?

Yep

charleskawczynski:
I think this is an upstream issue. I'm still not sure from where, but it might be productive to try using more recent versions to confirm if that's the case.

ilopezgp:
> I think this is an upstream issue. I'm still not sure from where, but it might be productive to try using more recent versions to confirm if that's the case.

Upstream from what packages? We should be up to date with all packages since the CompatHelper has not recommended any new PR.

charleskawczynski:
>Upstream from what packages? We should be up to date with all packages since the CompatHelper has not recommended any new PR.

Our `compat` entries allow for the latest version of SCS.jl, but we're not using the latest version of SCS.jl (the resolver gives us 0.9.something).

ilopezgp:
> > Upstream from what packages? We should be up to date with all packages since the CompatHelper has not recommended any new PR.
> 
> Our `compat` entries allow for the latest version of SCS.jl, but we're not using the latest version of SCS.jl (the resolver gives us 0.9.something).

Thanks @charleskawczynski !! I think we merged a compat change that did not test the changes from SCS. I tested SCS 0.9 and it leads to segfaults. SCS 1.0 cannot be used yet because of compat issues with Convex.jl, although I think they will be updating that package to v0.15 very soon. Version 0.8 works.

odunbar (approved):
Thanks this looks great! i think all my comments are now addressed. 

For the sake of codecov - could you add something to test the warnings that are thrown when 50% fail, or singular covariance? I think you could maybe see if the following are nonempty:
`@test_logs(:warn,) update_ensemble(...)` or
`@test_logs(:info,) update_ensemble(...)`

Then I'm happy to merge

ilopezgp:
> Thanks this looks great! i think all my comments are now addressed.
> 
> For the sake of codecov - could you add something to test the warnings that are thrown when 50% fail, or singular covariance? I think you could maybe see if the following are nonempty: `@test_logs(:warn,) update_ensemble(...)` or `@test_logs(:info,) update_ensemble(...)`
> 
> Then I'm happy to merge

Sounds good. I will add these as unit tests of the respective functions to be a bit cleaner.

Issue #113, opened 2022-03-02
Sparse operations in `SparseInversion` occur after updating the ekp object.


ilopezgp:
The SparseInversion update performs the following operations

```
# Sparse EKI
    cov_vv = vcat(hcat(cov_uu, cov_ug), hcat(cov_ug', cov_gg))
    H_u = hcat(1.0 * I(size(u)[1]), fill(FT(0), size(u)[1], size(g)[1]))
    H_g = hcat(fill(FT(0), size(g)[1], size(u)[1]), 1.0 * I(size(g)[1]))

    H_uc = H_u[ekp.process.uc_idx, :]

    cov_vv_inv = cov_vv \ (1.0 * I(size(cov_vv)[1]))

    # Loop over ensemble members to impose sparsity
    Threads.@threads for j in 1:(ekp.N_ens)
        # Solve a quadratic programming problem
        u[:, j] = sparse_qp(ekp, v[j, :], cov_vv_inv, H_u, H_g, y[:, j], H_uc = H_uc)

        # Threshold the results if needed
        if ekp.process.threshold_eki
            u[ekp.process.uc_idx, j] =
                u[ekp.process.uc_idx, j] .* (abs.(u[ekp.process.uc_idx, j]) .> ekp.process.threshold_value)
        end

        # Add small noise to constrained elements of u
        u[ekp.process.uc_idx, j] += rand(
            ekp.rng,
            MvNormal(zeros(size(ekp.process.uc_idx)[1]), ekp.process.reg * I(size(ekp.process.uc_idx)[1])),
        )
    end
```

after the ekp object has been updated,

```
push!(ekp.u, DataContainer(u, data_are_columns = true))
push!(ekp.g, DataContainer(g, data_are_columns = true))
```

This means that these manipulations do not really affect the advanced parameters. The initial block of operations copied in this issue should be done before pushing new DataContainers to the ekp object.

jinlong83:
It had no bug as it was consistent with previous version of DataContainer in terms of memory. I'll change the implementation accordingly to make it consistent with the updated version of DataContainer in #114

ilopezgp:
Fixed by #119

Issue #117, opened 2022-03-03
Issues with new versions of Convex and SCS


ilopezgp:
I tried updating to Convex 0.15 and SCS 1.0, and I ran into the warnings and errors below. In order to update, I followed the [release notes](https://github.com/jump-dev/Convex.jl/commit/4dde7383382631ef8712bb87eb154a7922e1621b) of Convex v0.15, which make the change

```
instead of `() -> SCS.Optimizer(verbose = 0)`, use `MOI.OptimizerWithAttributes(SCS.Optimizer, "verbose" => 0)`.
```

However, this leads to errors in the tests,

```

┌ Warning: Problem status ALMOST_OPTIMAL; solution may be inaccurate.
└ @ Convex ~/.julia/packages/Convex/o98db/src/solution.jl:342
┌ Warning: Problem status ALMOST_OPTIMAL; solution may be inaccurate.
└ @ Convex ~/.julia/packages/Convex/o98db/src/solution.jl:342
┌ Warning: Problem status ALMOST_OPTIMAL; solution may be inaccurate.
└ @ Convex ~/.julia/packages/Convex/o98db/src/solution.jl:342
┌ Warning: Problem status ALMOST_OPTIMAL; solution may be inaccurate.
└ @ Convex ~/.julia/packages/Convex/o98db/src/solution.jl:342
┌ Warning: Problem status ALMOST_OPTIMAL; solution may be inaccurate.
└ @ Convex ~/.julia/packages/Convex/o98db/src/solution.jl:342
┌ Warning: Problem status ALMOST_OPTIMAL; solution may be inaccurate.
└ @ Convex ~/.julia/packages/Convex/o98db/src/solution.jl:342
┌ Warning: Problem status ALMOST_OPTIMAL; solution may be inaccurate.
└ @ Convex ~/.julia/packages/Convex/o98db/src/solution.jl:342
┌ Warning: Problem status ALMOST_OPTIMAL; solution may be inaccurate.
└ @ Convex ~/.julia/packages/Convex/o98db/src/solution.jl:342
┌ Warning: Problem status ALMOST_OPTIMAL; solution may be inaccurate.
└ @ Convex ~/.julia/packages/Convex/o98db/src/solution.jl:342
┌ Warning: Problem status ALMOST_OPTIMAL; solution may be inaccurate.
└ @ Convex ~/.julia/packages/Convex/o98db/src/solution.jl:342
[...]

PR #119, opened 2022-03-03
Address issue #113

jinlong83:
The purposes of this pull request are:

- Update the implementation of sparse EKI to make it compatible with the new version of DataContainer.

### DIFF
--- a/src/SparseEnsembleKalmanInversion.jl
+++ b/src/SparseEnsembleKalmanInversion.jl
@@ -21,7 +21,7 @@ struct SparseInversion{FT <: AbstractFloat, IT <: Int} <: Process
 end
 
 function FailureHandler(process::SparseInversion, method::IgnoreFailures)
-    failsafe_update(u, g, y, obs_noise_cov, failed_ens) = eki_update(u, g, y, obs_noise_cov)
+    failsafe_update(ekp, u, g, y, obs_noise_cov, failed_ens) = sparse_eki_update(ekp, u, g, y, obs_noise_cov)
     return FailureHandler{SparseInversion, IgnoreFailures}(failsafe_update)
 end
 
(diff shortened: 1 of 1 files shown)


odunbar:
Looks good! could you merge base in? Then I'm happy to have this merged in

ilopezgp:
> Looks good! could you merge base in? Then I'm happy to have this merged in

I think this addresses #113 but not #117. Let's wait until both are resolved or change the name of the PR.

ilopezgp:
It might be a good idea to limit this PR to #113 so we can merge it quickly. I am not sure how involved the other one will be @jinlong83 .

src/SparseEnsembleKalmanInversion.jl
```diff
@@ -158,6 +154,10 @@ function update_ensemble!(
         )
     end
 
+    # store new parameters (and model outputs)
+    push!(ekp.u, DataContainer(u, data_are_columns = true))
+    push!(ekp.g, DataContainer(g, data_are_columns = true))
```
ilopezgp, line 159: Now that this part is moved after the sparsity modification, let's encapsulate L130-L155 along with the rest of update equations to a separate function called `sparse_eki_update`. You can see an example as `eki_update` in `EnsembleKalmanInversion.jl`. Then, let's call this function through the failsafe_update. I couldn't do it in the FailureHandler PR because the equations were split by these `push!` commands.

jinlong83:
> It might be a good idea to limit this PR to #113 so we can merge it quickly. I am not sure how involved the other one will be @jinlong83 .

I agree. I saw that Convex v0.14 has about 20 releases, and v0.15.0 just came out one day ago? I'd suggest stay with v0.14.x for now if possible, and I'll play with v0.15.x to better understand the issue of #117

src/SparseEnsembleKalmanInversion.jl
jinlong83: Done. Feel free to let me know if any other modification is needed.

ilopezgp (approved):

ilopezgp:
Thank you @jinlong83 ! Could you squash your commits before we merge? Thanks!

src/SparseEnsembleKalmanInversion.jl
```diff
@@ -88,7 +148,7 @@ function update_ensemble!(
     failed_ens = nothing,
 ) where {FT, IT}
 
-    # Update follows eqns. (4) and (5) of Schillings and Stuart (2017)
+    # Update follows eqns. (3.7) and (3.14) of Schneider et al. (2021)
```
ilopezgp, line 151: Minor comments, but you could move this comment to the `sparse_eki_update`.

ilopezgp:
I don't think you need to merge from master, just squash. Bors makes sure everything is ok before merging if there are no conflicts

PR #125, opened 2022-03-04
update cholesky in UKI

Zhengyu-Huang:
Update the Cholesky factor computation in UKI

When the matrix is not positive definite, compute SVD of the matrix, replace 0 singular values as the smallest nonzero singular value (with a threshold 1e-8), and then compute the Cholesky factor.

### DIFF
--- a/src/UnscentedKalmanInversion.jl
+++ b/src/UnscentedKalmanInversion.jl
@@ -231,7 +231,17 @@ function construct_sigma_ensemble(
 
     c_weights = process.c_weights
 
-    chol_xx_cov = cholesky(Hermitian(x_cov)).L
+    # compute cholesky factor L of x_cov
+    local chol_xx_cov
+    try
+        chol_xx_cov = cholesky(Hermitian(x_cov)).L
+    catch
+        _, S, Ut = svd(x_cov)
+        # find the first singular value that is smaller than 1e-8
+        ind_0 = searchsortedfirst(S, 1e-8, rev = true)
+        S[ind_0:end] .= S[ind_0 - 1]
+        chol_xx_cov = (qr(sqrt.(S) .* Ut).R)'
+    end
 
     x = zeros(FT, N_x, 2 * N_x + 1)
     x[:, 1] = x_mean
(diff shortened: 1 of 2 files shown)


src/UnscentedKalmanInversion.jl
```diff
@@ -172,7 +172,16 @@ function construct_sigma_ensemble(
 
     c_weights = process.c_weights
 
-    chol_xx_cov = cholesky(Hermitian(x_cov)).L
+    # compute cholesky factor L of x_cov
+    try
+        global chol_xx_cov = cholesky(Hermitian(x_cov)).L
```
ilopezgp, line 177: Is it safe to use `global` here @charleskawczynski ?

ilopezgp:
Can you pass a unit test of `construct_sigma_ensemble` passing a positive semidefinite covariance matrix?

src/UnscentedKalmanInversion.jl
Zhengyu-Huang: I will add a unit test, if I do not put global here, I get an error. the scope of try and catch is weird

charleskawczynski, line 177:
```suggestion
    local chol_xx_cov
    try
        chol_xx_cov = cholesky(Hermitian(x_cov)).L
```

src/UnscentedKalmanInversion.jl
```diff
@@ -172,7 +172,16 @@ function construct_sigma_ensemble(
...
+    catch
+        _, S, Ut = svd(x_cov)
+        # find the first singular value that is smaller than 1e-8
+        ind_0 = searchsortedfirst(S, 1e-8, rev = true)
+        S[ind_0:end] .= S[ind_0 - 1]
+        global chol_xx_cov = (qr(sqrt.(S) .* Ut).R)'
```
charleskawczynski, line 183:
```suggestion
       chol_xx_cov = (qr(sqrt.(S) .* Ut).R)'
```

src/UnscentedKalmanInversion.jl
charleskawczynski: I would typically use `local` here, maybe global works the same way? I think `local` is more intuitive.

Zhengyu-Huang: just update to `local` and add a test

ilopezgp (approved):
LGTM, thank you @Zhengyu-Huang !

ilopezgp:
Closes #126

PR #133, opened 2022-03-28
Simplify SparseInversion struct, add more constructors.

ilopezgp:
- Removes `threshold_eki` boolean, since it can be inferred from magnitude of `threshold_value`.
- Sets `reg` last, since it is a parameter only needed for numerical stability and ideally not user-facing. 
-  Verifies that `reg` is positive definite before attempting to sample from a gaussian with `reg` as its variance.
- Adds base kwarg constructor to SparseInversion.
- Adds defaults to struct properties and an outer constructor that accepts a flexible number of keyword parameters.

### DIFF
--- a/src/SparseEnsembleKalmanInversion.jl
+++ b/src/SparseEnsembleKalmanInversion.jl
@@ -7,17 +7,24 @@ using SparseArrays
 
 A sparse ensemble Kalman Inversion process
 """
-struct SparseInversion{FT <: AbstractFloat, IT <: Int} <: Process
+Base.@kwdef struct SparseInversion{FT <: AbstractFloat} <: Process
     "upper limit of l1-norm"
     γ::FT
-    "the flag to enable thresholding"
-    threshold_eki::Bool
-    "a small value to threshold results that are close to zero"
-    threshold_value::FT
-    "a small value for regularization to enhance robustness of convex optimization"
-    reg::FT
-    "a list of index to indicate the parameters included in the evaluation of l1-norm"
-    uc_idx::AbstractVector{IT}
+    "threshold below which the norm of parameters is pruned to zero"
+    threshold_value::FT = FT(0)
+    "indices of parameters included in the evaluation of l1-norm constraint"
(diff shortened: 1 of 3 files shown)


src/SparseEnsembleKalmanInversion.jl
```diff
@@ -115,24 +122,22 @@ function sparse_eki_update(
...
+        # Prune parameters using threshold
+        u[ekp.process.uc_idx, j] =
+            u[ekp.process.uc_idx, j] .* (abs.(u[ekp.process.uc_idx, j]) .> ekp.process.threshold_value)
 
         # Add small noise to constrained elements of u
-        u[ekp.process.uc_idx, j] += rand(
-            ekp.rng,
-            MvNormal(zeros(size(ekp.process.uc_idx)[1]), ekp.process.reg * I(size(ekp.process.uc_idx)[1])),
-        )
+        if isposdef(ekp.process.reg)
```
jinlong83, line 130: Is it a convention for our codes to use `isposdef()` for scalars?

src/SparseEnsembleKalmanInversion.jl
```diff
@@ -115,24 +122,22 @@ function sparse_eki_update(
...
+            sample_size = length(u[ekp.process.uc_idx, j])
```
jinlong83, line 131: The name `sample_size` might be a little bit ambiguous (what we try to evaluate here is the number of sparse elements).

src/SparseEnsembleKalmanInversion.jl
ilopezgp: Not really, I am using it here because it guarantees the same precision needed to draw elements from a distribution (there is an internal isposdef check to the covariance matrix)

src/SparseEnsembleKalmanInversion.jl
ilopezgp: I will change it to `len_u_sparse`

jinlong83: Thanks! The other edits look good to me. Please feel free to merge this PR.

PR #138, opened 2022-03-31
New parameter Constructor

### LINKED ISSUE #131: Cleaning up ParameterDistribution constructor
Due to my coding naivity, we have things that look like this
https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/966bec136ce37d1bf14a888be931f5f8f1f1a519/examples/LossMinimization/loss_minimization.jl#L41-L47
Where we see constraints are always arrays, or other oddities. 

We should instead ask that we create these with a non-flattened format initially e.g. array of dictionaries 
```julia
prior_u = 
[
    Dict(
        name => "u1", 
        distribution => Parameterized(Normal(0, 1)), 
        constraints => no_constraint()
    ),
    Dict(
        name => "u2", 
        distribution => Parameterized(MvNormal(4, 0.1)), 
        constraints => [no_constraint(), no_constraint(), no_constraint(), no_constraint()]
    )
]
u = ParameterDistribution(prior_u)
```
I'm willing to iterate on precise forms here. But something along these lines will be far easier to do all the dimensionality checks while in this structured form, before we flatten it internally.

odunbar:
## Purpose

Resolves #131 , in particular we should be free of having to define constraints in specific array shapes.

## In this PR
- Modification of the original constructor to push for build single distribution, and a function `combine_distributions` to merge an array of distributions
- A new convenience constructor for a `ParameterDistribution` object. Constructing from a dictionary, and resolving the issue with constraints needing to be arrays.
- Unit tests for the new checks in the constructor(s)
- All EKP examples now use `Dict` constructor
- New documentation for the prior distributions page
### From `Cloudy/Cloudy_example_eki.jl`: Modified Constructor way
```julia
lbound_N0 = 0.4 * N0_true
prior_N0 = ParameterDistribution(
    Parameterized(Normal(4.5, 1.0)), #truth is 5.19
    bounded_below(lbound_N0),
    "N0",
)
lbound_θ = 1.0e-1
prior_θ = ParameterDistribution(
    Parameterized(Normal(0.0, 2.0)),  #truth is 0.378
    bounded_below(lbound_θ),
    "θ",
)
lbound_k = 1.0e-4
prior_k = ParameterDistribution(
    Parameterized(Normal(-1.0, 1.0)), #truth is -2.51 
    bounded_below(lbound_k),
    "k",
)
priors = combine_distributions([u1,u2,u3])
```
### From `Cloudy/Cloudy_example_eki.jl`: Dictionary way
 Parameter distributions are built individually as dicts and then combined in an array.
```julia 
# N0
lbound_N0 = 0.4 * N0_true
prior_N0 = Dict(
    "distribution" => Parameterized(Normal(4.5, 1.0)), #truth is 5.19
    "constraints" => bounded_below(lbound_N0),
    "name" => "N0",
)
# θ
lbound_θ = 1.0e-1
prior_θ = Dict(
    "distribution" => Parameterized(Normal(0.0, 2.0)),  #truth is 0.378
    "constraints" => bounded_below(lbound_θ),
    "name" => "θ",
)
# k
lbound_k = 1.0e-4
prior_k = Dict(
    "distribution" => Parameterized(Normal(-1.0, 1.0)), #truth is -2.51 
    "constraints" => bounded_below(lbound_k),
    "name" => "k",
)

priors = ParameterDistribution( [prior_N0, prior_θ, prior_k] )
```


bielim:
I agree that the old constructor's inconsistent treatment of constraints and distributions (constraints always had to be packed into arrays, whereas distributions could be single objects) was a source of confusion, so I welcome the development of a more consistent constructor (even though it will cause my UQ parameter file parser to break :-)).

A few comments:

- I was thinking that in terms of the data type that hold the parameter information and gets passed as argument to the constructor, structs might be a better (more efficient) choice than dictionaries since we are dealing with a fixed amount of data of known type (three pieces of data for each parameter: distribution, constraint, name). Constructing structs also requires less writing on the part of the user since you can directly pass the arguments without having to write the keys -- "distribution", "constraints", "name" --- for each parameter. E.g.
```
struct ParamInfo
    distribution::ParameterDistributionType
    constraint::ConstraintType
    name::AbstractString
end

param_info = ParamInfo(Parameterized(Gamma(2.0, 3.0)), no_constraint(), "foo")
```
- I can also see the dictionaries making the "broadcasting functionality" we implemented for UQ parameters more tedious,  if you think e.g. of a case where you have to generate 100 individual dictionaries, each with a key "name" whose value is derived from the corresponding table header in the .toml file (e.g., uq_param_(1), ...., uq_param_(100)). If you're using structs you'd admittedly still have to generate 100 of them, but the advantage is that you can construct an array of structs by separately defining an array of distributions, an array of constraints, and an array of names, and then use `StructArrays.jl` to do
```
# distributions is an array of 100 distributions, 
# constraints is an array of 100 constraints,
# names is an array of 100 names derived from the parameter's "base name"
array_of_structs = StructArray(ParamInfo(distributions[i], constraints[i], names[i]) for i in 1:100)
u = ParameterDistribution(array_of_structs) 
```
- One thing that I'm not sure how best to address (or maybe it doesn't need to be addressed at all), but that has caused a bit of trouble for the parsing as well: When you describe a number of parameters as a multivariate distribution, the `ParameterDistribution` only requires a single parameter name, so e.g. a distribution `Parameterized(MvNormal(4, 1.0))` can be given the name `"my_4dim_parameter"`, whereas if you decide to represent the distribution as four separate univariate distributions (since the dimensions are independent), you would have to provide four names. How about if we always required one parameter name per dimension instead?
- Finally, I would for the sake of consistency use either singular or plural for all pieces of information ("distribution", "constraint", and "name", or "distributions", "constraints", and "names")

odunbar:
@bielim Thanks for the comments!

1. Though we could use structs, I would want to use them based on some needing to store type information, or if we wanted to construct `ParamInfo` with different constructors. I used Dict's simply as a way of swapping order for keywords. Your suggestion seems to be like creating a simple `ParameterDistribution` but with a different name?

To that end we could also add a new constructor to  `ParameterDistribution`
```julia
ParameterDistribution = ParameterDistribution(array_of_ParameterDistributions)
```
So you can choose to build a parameter distribution from an array of "simpler" parameter distributions? 

2. Not sure i know what you mean with the broadcasting as that can be done easily with dicts too:
```julia
dict_array = [ Dict("distribution" => distributions[i], "constraints" => constraints[i], "name" => names[i]) for i in 1:100]
u = ParameterDistribution(dict_array) 
```

3. From the point of the clima application, (correct me if I'm wrong) I think one name per distribution is correct, because the names are determined by their use in the clima code. The only oddity i see at the moment is the special "repeat" case you've developed, which will not work in the proposed frameworks, without repeating the name too. 
 
6. We can try making everything singular.

bielim:
> 1. Though we could use structs, I would want to use them based on some needing to store type information, or if we wanted to construct `ParamInfo` with different constructors. I used Dict's simply as a way of swapping order for keywords. Your suggestion seems to be like creating a simple `ParameterDistribution` but with a different name?

Why do we need the ability to swap the order for keywords? Structs are more efficient than dictionaries, and in a case like this one, where you know the types of arguments in advance, it seems that the flexibility dictionaries provide (which comes at an efficiency cost) is not needed, and that explicitly specifying these types would make things more explicit and clear. I mean, since this code is hardly ever going to be a computational bottleneck, efficiency gains are not that crucial, I just think that structs are the more appropriate choice of data container in this case, plus they allow for definition of custom constructors, which may turn out to be convenient down the road.
I don't quite see how my suggestion is like creating a simple `ParameterDistribution` but with a different name, or if it is, then I don't see why the version with the dictionary is not :-) (I mean, both the dictionary and the struct simply package the information about the parameter before it gets passed to the `ParameterDistribution` constructor, so one could argue that it is an unnecessary step compared to the `ParameterDistribution` constructor that directly takes the "raw" parameter information, which is what the original implementation did.)


> 3\. Not sure i know what you mean with the broadcasting as that can be done easily with dicts too:
> 
> ```julia
> dict_array = [ Dict("distribution" => distributions[i], "constraints" => constraints[i], "name" => names[i]) for i in 1:100]
> u = ParameterDistribution(dict_array) 
> ```

Yeah that's true (I like not having to write out the keys better, but that's just a cosmetic problem)

> 3\. From the point of the clima application, (correct me if I'm wrong) I think one name per distribution is correct, because the names are determined by their use in the clima code. The only oddity i see at the moment is the special "repeat" case you've developed, which will not work in the proposed frameworks, without repeating the name too. I'll have a look at this.

The "oddity" I can see in the example I described is that these are two equivalent ways of specifying the same distribution, but one requires 4 parameter names and the other one requires only 1 name. But I should probably clarify that I don't mind having to generate "numbered parameter names" for these "repeat" cases (since after all the user __is__ generating N distributions and constraints for N parameters, so it seems natural that each parameter would get its own name). I just noticed that, bei using multivariate (N-dimensional) distributions, you can end up constructing cases that are equivalent to N univariate distributions, but the requirements for the number of names differ between the two cases. As I said, it's more of an observation, not sure if there is something to be done here.



> 4\. We can try making everything singular.

:+1:

odunbar:
Let's take the discussion of 1) and 2) offline. I am certainly not against structs.

src/ParameterDistributions.jl
```diff
@@ -160,39 +161,76 @@ n_samples(d::Parameterized) = "Distribution stored in Parameterized form, draw s
...
         end
     end
 
-end
+    #make copy as array
+    param_dist_dict_array = !isa(param_dist_dict, AbstractVector) ? [param_dist_dict] : param_dist_dict
+    #perform checks on the individual distributions
+    for pdd in param_dist_dict_array
+        #check all keys are present
+        if !all(k in collect(keys(pdd)) for k in ["distribution", "name", "constraints"])
```
haakon-e, line 191:
```suggestion
        if !all(keys(pdd) .∈ [["distribution", "name", "constraints"]])
```

src/ParameterDistributions.jl
```diff
@@ -160,39 +161,76 @@ n_samples(d::Parameterized) = "Distribution stored in Parameterized form, draw s
...
+"""
+function ParameterDistribution(param_dist_dict::Union{Dict, AbstractVector})
 
-            new{PDType, ConstraintType, ST}(parameter_distributions, constraints, names)
+    #check type
+    if !isa(param_dist_dict, Dict)
+        if !(eltype(param_dist_dict) <: Dict)
+            throw(ArgumentError("input argument must be a Dict, or <:AbstractVector{Dict}"))
         end
     end
```
haakon-e, line 184:
```suggestion
    if !(isa(param_dist_dict, Dict) || eltype(param_dist_dict) <: Dict)
        throw(ArgumentError("input argument must be a Dict, or <:AbstractVector{Dict}"))
    end
```

src/ParameterDistributions.jl
```diff
@@ -160,39 +161,76 @@ n_samples(d::Parameterized) = "Distribution stored in Parameterized form, draw s
...
+        if !isa(distribution, ParameterDistributionType)
+            throw(ArgumentError("Value of \"distribution\" must be a valid ParameterDistribution object: Parameterized or Samples"))
+        end
+        if !isa(constraints, ConstraintType)
+            if !isa(constraints, AbstractVector) #it's not a vector either
+                throw(ArgumentError("Value of \"constraints\" must be a ConstraintType, or <:AbstractVector(ConstraintType)"))
+            elseif !(eltype(constraints) <: ConstraintType) #it is a vector, but not of constraints
+                throw(ArgumentError("Value of \"constraints\" must be a ConstraintType, or <:AbstractVector(ConstraintType)"))
+            end
+
```
haakon-e, line 209:
```suggestion
```

src/ParameterDistributions.jl
```diff
@@ -160,39 +161,76 @@ n_samples(d::Parameterized) = "Distribution stored in Parameterized form, draw s
...
 
+        if !(n_parameters == length(constraints_array))
+            throw(DimensionMismatch("There must be one constraint dimension in a parameter distribution, use no_constraint() type if no constraint is required"))
+        end
+    end
+
+    #flatten the structure
+    distributions = [pdd["distribution"] for pdd in param_dist_dict_array]
+    flat_constraints = cat([pdd["constraints"] for pdd in param_dist_dict_array]..., dims = 1)
+    names = [pdd["name"] for pdd in param_dist_dict_array]
```
haakon-e, line 228:
```suggestion
    distributions = getindex.(param_dist_dict_array, "distribution")
    flat_constraints = cat(getindex.(param_dist_dict_array, "constraints")..., dims = 1)
    names = getindex.(param_dist_dict_array, "name")
```

haakon-e (commented):
Just a few (mostly stylistic) code suggestions for clarity. Looks good overall. I like this way of grouping information.

src/ParameterDistributions.jl
odunbar: Done

src/ParameterDistributions.jl
odunbar: done

src/ParameterDistributions.jl
odunbar: done

src/ParameterDistributions.jl
odunbar: I think this should be the other way around, we check if "[distribution]... form a subset of keys(pdd)" rather than the other way around. But otherwise i can compact it.

odunbar: done

docs/src/parameter_distributions.md
```diff
@@ -3,11 +3,15 @@
 We provide a flexible setup for storing prior distribution with the `ParameterDistributions` module found in `src/ParameterDistributions.jl`.
 
 One can create a full parameter distribution using three inputs:
- 1. A Distribution, given as a `ParameterDistributionType` object,
- 2. An array of Constraints, given as a `Array{ConstraintType}` object,
- 3. A Name, given as a `String`.
+ 1. A distribution, given as a `ParameterDistributionType` object,
+ 2. A constraint, or array of constraints, given as a `Array{ConstraintType}` object,
```
ilopezgp, line 7:
```suggestion
 2. A constraint, or array of constraints, given as a `ConstraintType` or `Array{ConstraintType}` object,
```

ilopezgp (approved):
LGTM

docs/src/parameter_distributions.md
odunbar: Done

odunbar:
@ilopezgp @haakon-e @bielim @costachris @tsj5 Please note the alternative construction added based on modified constructors. 
Both dicts and constructor methods have a similar mantra of "build simple objects and combine". I'd be happy with either (or both) options, but would like opinions on whether we should remove any to give users a single way of building or to keep both, 

Cheers!

bielim:
I prefer the alternative construction (https://github.com/CliMA/EnsembleKalmanProcesses.jl/issues/131#issuecomment-1092712097) and would vote for giving users a single way of building parameter distribution objects.

ilopezgp:
I prefer the constructor that uses a dictionary or array of dictionaries, so why don't we leave both?

odunbar:
OK - I will leave both for now. If it becomes burdensome to maintain we can re-evaluate down the line. 

I'll merge later today.

odunbar:
bors try

PR #141, opened 2022-04-08
[WIP] Constructors for priors with constraints

tsj5:
It's desirable to streamline the UI for specifying parameter priors, in particular those for parameters that live on a constrained domain and hence need a nontrivial Constraint().

This PR proposes to do this by adding new constructors for ParameterizedDistribution. The unconstrained distribution is a Gaussian, with coefficients of the constrained-unconstrained transformation fit at runtime so that the constrained distribution meets user specifications:

- The interval the constrained distribution is defined on. One or both interval endpoints may be infinite.
- A metric of the prior expected value of the parameter (either mean or median).
- A metric of the spread of the prior distribution (standard deviation, inter-quartile or inter-quantile range).

Emphasis is given to quantiles because, in these cases, the transformation may be obtained in closed form (see doc). For the more relevant mean/standard deviation cases, we need to optimize the coefficients numerically, which we plan to do using the quantile-based solution as an efficient initial guess.


tsj5:
Withdrawing in response to feedback from @odunbar. Replaced by #161 .

PR #145, opened 2022-04-11
[WIP] VectorOfParameterized distribution type

### LINKED ISSUE #142: Require new parameter type for "vector of parameterized" distributions
## Issue

Currently `Parameterized` has limitations to types that Julia specified. A new type that appears in practice e.g. for specifying independent priors for a function at 100 points, we may wish to define
```julia
distribution = repeat([Parameterized(Normal(0,1))],100)
constraint = repeat([no_constraint()],100)
name = "function_at_100_points"
```
Our current setup can do either 
1. replace `repeat([Parameterized(Normal(0,1))],100)` with a single `Parameterized(MvNormal(100,1.0))`. This is restrictive to only Multivariate distributions from `Distributions`
2. replace name with the array of names `name = [string("function_at_point_",i) for i =1:100]`. This means that each of the 100 distributions have different names, and may require additional parsing.

## Proposal: `VectorOfParameterized` distribution type
```julia 
distribution = VectorOfParameterized( repeat([Normal(0,1)],100) )
constraint = repeat([no_constraint()],100)
name = "function_at_100_points"
```
when provided with corresponding functions sample/var/cov etc. this fits within the current framework. The stacked distributions can have different priors and be different dimensions.

odunbar:
## Purpose
Resolves #142 

## In the PR
- New `VectorOfParameterized` distribution struct. Can create a single distribution from an array of (possibly different, univariate and multivariate) Distributions from `Distributions.jl` 
- Accompanying sampling/statistical/shape functions
- Unit tests
- Docs in parameter distributions
## Also
- removed a bug where input dictionary is modified
## Simple example
Creating a full `ParameterDistribution` from 100 univariate normals (e.g. often used in  parameters)
```julia
distribution = VectorOfParameterized(repeat([Normal(0,1)],100)
constraint = repeat([no_constraint()],100)
name = "100 normals"
ParameterDistribution(distribution, constraint, name)
```
Key observation: only one name is required for the distribution vector - this for example is a useful addition for parsing array-style parameterrs e.g. the representation and discretization of functions.


bielim (approved):
This looks good to me!

The only thing I noticed was something that was introduced in a previous PR (sorry for mentioning it here - provided that we decide this line should be modified, I can also open a separate PR for it if needed): 
Line 229:
`pdd["constraint"] = constraint_array # make the copy constraint always an array`
This silently modifies the argument (`param_dist_dict`, in particular the value of "constraint") of the `ParameterDistribution`, which may lead to confusion if the user wants to use the parameter dictionaries (or parts of them) for some other purpose after passing them to the `ParameterDistribution` constructor. Would there be any damage from just deleting that line?

odunbar:
@bielim good spot - I have removed the line, I think when I first wrote the constructor i made a copy at the top of pdd just in case. but since was removed! merging now...

PR #161, opened 2022-05-17
Constructors for Gaussian priors with constraints

tsj5:
It's desirable to streamline the UI for specifying parameter priors, in particular those for parameters that live on a constrained domain and hence need a nontrivial Constraint().

This PR does this by adding a new constructor, ~~bounded_gaussian~~ `constrained_gaussian`, for ParameterizedDistribution. The unconstrained distribution is a Gaussian, with parameters (μ_u, σ_u) selected to reproduce user-required statistics in the constrained space [currently only (mean, standard deviation) are supported]. This therefore includes the [lognormal](https://en.wikipedia.org/wiki/Log-normal_distribution) and [logit-normal](https://en.wikipedia.org/wiki/Logit-normal_distribution) distributions as special cases.

In the cases where the constraint interval is infinite or semi-infinite, closed-form expressions are used for (μ_u, σ_u). When the interval is finite, no closed-form solution exists, and (μ_u, σ_u) are optimized numerically via the simplex method. This is still less reliable than it should be for large |μ_u| or σ_u: for these cases the distribution becomes concentrated near one or both interval endpoints, causing instabilities for both numerical integration and optimization (since parameter changes have minimal effect on the objective function).

This PR fixes a sign error in bounded_above() which caused its Jacobian to be negative.

Tasks remaining as of this writing:
- constructor for distributions given as Cartesian products of the 1D case currently implemented.
- interface for using `constrained_gaussian` to specify priors from parameter file conventions.


tsj5:
Does anyone know how to fix the failing JuliaFormatter/format [test](https://github.com/CliMA/EnsembleKalmanProcesses.jl/runs/6466474003?check_suite_focus=true)? I've run `julia .dev/climaformat.jl` locally, and running it again doesn't change any files. Is the version of JuliaFormatter that's run in CI using different settings than climaformat.jl?

src/ParameterDistributions.jl
```diff
@@ -667,4 +694,145 @@ function transform_unconstrained_to_constrained(
...
+    with adequate data, so for the sake of performance we only require that the optimization
+    reproduce μ_c, σ_c to a loose tolerance (1e-5). Warnings are logged when the optimization
+    fails.
+
+!!! note
+    The distribution may be bimodal for σ_c large relative to the width of the bound interval.
+    In extreme cases the distribution becomes concentrated at the bound endpoints. We regard
+    this as a feature, not a bug, and do not warn the user when bimodality occurs.
+"""
+function bounded_gaussian(name::AbstractString, μ_c::FT, σ_c::FT, lower_bound::FT, upper_bound::FT) where {FT <: Real}
```
odunbar, line 748: Maybe the name should not be `bounded` as it is not always bounded, and maybe it should allude to the fact that it is a constructor for a `ParameterDistribution` object. How about `create_constrained_gaussian_distribution`? Or something along these lines.

src/ParameterDistributions.jl
```diff
@@ -667,4 +694,145 @@ function transform_unconstrained_to_constrained(
...
+    init_logσ_u = log(init_σ_u)
+
+    # Optimize parameters; could speed up with gradient info but don't bother to.
+    # Optimization is finicky since problem becomes singular for large |μ_u|, σ_u: the 
+    # distribution becomes collapsed against the bounds of the interval.
+    function _optim_fn(μlogσ::Vector{Float64})
+        m, s = _mean_std(μlogσ[1], exp(μlogσ[2]), cons)
+        return (m - μ_c)^2 + (s - σ_c)^2
+    end
+    opt = optimize(_optim_fn, [init_μ_u, init_logσ_u], NelderMead(), Optim.Options(x_tol = 1e-5, f_tol = 1e-5))
```
odunbar, line 824: I think if you are calling solvers internally - it would be sensible to have a `optim_kwargs` type arguments at the top level?

src/ParameterDistributions.jl
```diff
@@ -667,4 +694,145 @@ function transform_unconstrained_to_constrained(
...
+    μ_u = Optim.minimizer(opt)[1]
+    σ_u = exp(Optim.minimizer(opt)[2])
+
+    m_c, s_c = _mean_std(μ_u, σ_u, cons)
+    if ~isapprox(μ_c, m_c, atol = 1e-3, rtol = 1e-2)
+        @warn "Unable to set constrained mean for `$(name)`: target = $(μ_c), got $(m_c)"
+    end
+    if ~isapprox(σ_c, s_c, atol = 1e-3, rtol = 1e-2)
```
odunbar, line 832: If these are taken in the constrained space, perhaps they should be normalized? (or is that what rtol does?)

src/ParameterDistributions.jl
```diff
@@ -667,4 +694,145 @@ function transform_unconstrained_to_constrained(
...
+    # Numerically estimate μ_u, σ_u for unbounded distribution which reproduce desired
+    # μ_c, σ_c in constrained, transformed coordinates. Unlike other cases, can't be solved
+    # for analytically.
+
+    cons = bounded(lower_bound, upper_bound)
+    init_μ_u, init_σ_u = _bounded_gaussian_guess(μ_c, σ_c, lower_bound, upper_bound)
+    # optimize in log σ to avoid constraint σ>0
+    init_logσ_u = log(init_σ_u)
+
+    # Optimize parameters; could speed up with gradient info but don't bother to.
```
odunbar, line 817: If we don't need gradients then they can be removed from the constraints?

odunbar (commented):
Hi Tom - this looks much more along the lines of what I was looking for.
In the todo's, could you also add something to the docs to show how it works?

tsj5:
Resolution of the climaformat issue mentioned [above](https://github.com/CliMA/EnsembleKalmanProcesses.jl/pull/161#issuecomment-1129294087), written up here for discoverability: @odunbar mentioned that the version of JuliaFormatter had been updated to 0.22; this was in .dev/Project.toml but for some reason .dev/Manifest.toml was still using the older version (I would have thought that that would've been caught and updated when that environment was `activate`'ed). Deleting the Manifest and re-instantiating solved the issue.

src/ParameterDistributions.jl
tsj5: Sorry, I'm lost on the proper context for "normalized" -- normalization, svd. etc. are only done in CES, right? 

`m_c` and `s_c` are the mean/stddev of the distribution at its final, hopefully optimized parameter values, computed as 1d numerical integrals in the physical, constrained space. The highlighted lines are a sanity check to verify that the optimization set those parameters to values that satisfy the user's requested `μ_c`, `σ_c` for the prior, to very loose tolerances. `rtol` is just the tolerance on the relative difference between the two values.

src/ParameterDistributions.jl
tsj5: Is `constrained_gaussian` OK? (i.e. 'create' and 'distribution' being clear from context). climaformat limits us to 120 character lines, and there's no explicit line continuation marker in Julia.

odunbar (commented):
As this is user facing - could you also provide an example snippet and some information in the docs priors page?

Otherwise, LGTM!

tsj5:
Finally added docs and examples for `constrained_gaussian`. See GitHub's markdown preview [here](https://github.com/tsj5/EnsembleKalmanProcesses.jl/blob/tsj/prior-init-2/docs/src/parameter_distributions.md) -- I don't think there's any way to trigger the docs build for PRs, right?

docs/src/parameter_distributions.md
```diff
@@ -21,119 +26,106 @@ dict_1 = Dict("distribution" => distribution_1, "constraint" => constraint_1, "n
 dict_2 = Dict("distribution" => distribution_2, "constraint" => constraint_2, "name" => name_2)
 prior = ParameterDistribution( [dict_1, dict_2] )
 ```
+
 We provide many examples in the package `examples/`, the unit tests found in `test/ParameterDistributions/runtests.jl`, and in the documentation below.
 
-# A simple example:
-Task: We wish to create a prior for a one-dimensional parameter. Our problem dictates that this parameter is bounded between 0 and 1. Prior knowledge dictates it is around 0.7. The parameter is called `point_seven`.
+### Constrained\_gaussian constructor
```
odunbar, line 32: Perhaps reword? "The `constrained_gaussian` constructor (recommended)"

docs/src/parameter_distributions.md
```diff
@@ -21,119 +26,106 @@ dict_1 = Dict("distribution" => distribution_1, "constraint" => constraint_1, "n
...
 
-Solution: We should use a Normal distribution with the predefined "bounded" constraint.
+The constructors described above can be used to specify arbitrary prior distributions and constraints. However, the anticipated use case for EnsembleKalmanProcesses is one where inference is not strongly prior-dominated: i.e., the user only needs (or is only able to) specify the physical and likely ranges of prior parameter values at a rough, qualitative level.
```
odunbar, line 34: `EnsembleKalmanProcesses`? in quotes?

odunbar: Remove the phrase "strongly prior dominated", perhaps instead just say "where prior information is qualitative, and exact distributions of the priors are unknown."

docs/src/parameter_distributions.md
```diff
@@ -21,119 +26,106 @@ dict_1 = Dict("distribution" => distribution_1, "constraint" => constraint_1, "n
...
+We provide the following predefined constructors which implement mappings that handle the most common constraints:
 
- - `no_constraint()`, no transform is required for this parameter,
- - `bounded_below(lower_bound)`, the physical parameter has a (provided) lower bound,
- - `bounded_above(upper_bound)`, the physical parameter has a (provided) upper bound,
- - `bounded(lower_bound,upper_bound)`, the physical parameter has the (provided) bounds.
+ - `no_constraint()`: The parameter is unconstrained and takes values in (-∞, ∞) (mapping is the identity).
+ - `bounded_below(lower_bound)`: The parameter takes values in [`lower_bound`, ∞).
+ - `bounded_above(upper_bound)`: The parameter takes values in (-∞, `upper_bound`].
+ - `bounded(lower_bound,upper_bound)`: The parameter takes values on the interval [`lower_bound`, `upper_bound`].
```
odunbar, line 78: I like this - thanks!

docs/src/parameter_distributions.md
```diff
@@ -1,13 +1,18 @@
...
+
+ 1. The distribution itself, given as a [ParameterDistributionType](@ref) object. This includes standard Julia [Distributions](https://juliastats.org/Distributions.jl/stable/) as well as empirical/sample-based distributions, and can be uni- or multi-variate. To clarify, despite our use of the term "Kalman processes," the prior distribution is *not* required to be Gaussian.
+ 2. A constraint (or array of constraints) on the domain of the distribution, given as a [ConstraintType](@ref) or `Array{ConstraintType}` object (the latter case builds a multivariate constraint as the Cartesian product of one-dimensional Constraints). This is used to enforce physical parameter values during inference: the model is never evaluated at parameter values outside the constrained region, and the posterior distribution will only be supported there.
+ 3. The parameter name, given as a `String`.
+
+In multiparameter settings, one should define one `ParameterDistribution` per parameter, and then concatenate these either in the constructor or with `combine_distributions`. This is illustrated below and in the [Multidimensional example](@ref).
+
+### General constructor
```
odunbar, line 15: Do you think we could move the General constructor below - so the recommended constructor comes up first?

docs/src/parameter_distributions.md
```diff
@@ -21,119 +26,106 @@ dict_1 = Dict("distribution" => distribution_1, "constraint" => constraint_1, "n
...
+prior_1 = constrained_gaussian("param_name", μ_1, σ_1, 0.0, Inf)
+prior_2 = constrained_gaussian("param_name", μ_2, σ_2, lower_bound, upper_bound)
+prior = combine_distributions( [prior_1, prior_2])
 ```
 
-```@setup zero_point_seven
-using Distributions
-using Plots
-Plots.default(lw=2)
+This constructs a prior distribution on the interval [`lower_bound`, `upper_bound`] with mean μ and standard deviation σ. The distribution itself is a Guassian "squashed" according to the interval chosen (e.g. `prior_1` is a log-normal.) See [Simple example - 1](@ref) for an example.
```
odunbar, line 44: Typo "Guassian"

docs/src/parameter_distributions.md
```diff
@@ -361,28 +357,108 @@ end
 gif(anim_bounded, "anim_bounded.gif", fps = 10) # hide
 ```
 
-The following generates the transformed `Normal(0.5, 1)` distribution
 
+
+## Full Examples
+
+### Simple example - 1
```
odunbar, line 364: Once again, I think rearranging to put the recommended constructor first would be good here

docs/src/parameter_distributions.md
```diff
@@ -361,28 +357,108 @@ end
...
+vline!([0.7]) 
+title!("Prior pdf")
+```
+
+The pdf of the constructed prior distribution (in the physical, constrained space) looks like:
+```@example example_one
+p = plot(p2, legend=false, size = (450, 450)) #hide
+```
+
+### Simple example - 2
```
odunbar, line 402: It would be great if these "simple examples" were moved above all the constraints discussion

odunbar (commented):
Sorry it's taken a while for review - this looks great. Please merge after addressing my comments.
- My only real point is that now we have a "recommended" constructor, we should present this first, likewise with the simple example for it. (Most users will only have to read the first part of this page). After we can talk about general constructors, and all the details for the constraints etc.

docs/src/parameter_distributions.md
tsj5: Addressed in 5ef7856

docs/src/parameter_distributions.md
tsj5: Addressed in [9b09cea](https://github.com/CliMA/EnsembleKalmanProcesses.jl/pull/161/commits/9b09cea7032cefb92cd0332b9656b924152356c1)

docs/src/parameter_distributions.md
tsj5: Good catch! Fixed in [9b09cea](https://github.com/CliMA/EnsembleKalmanProcesses.jl/pull/161/commits/9b09cea7032cefb92cd0332b9656b924152356c1)

docs/src/parameter_distributions.md
tsj5: Done in [9b09cea](https://github.com/CliMA/EnsembleKalmanProcesses.jl/pull/161/commits/9b09cea7032cefb92cd0332b9656b924152356c1)

docs/src/parameter_distributions.md
tsj5: Addressed in [9b09cea](https://github.com/CliMA/EnsembleKalmanProcesses.jl/pull/161/commits/9b09cea7032cefb92cd0332b9656b924152356c1), which implements the recommended reorganization.

tsj5:
> Sorry it's taken a while for review - this looks great. Please merge after addressing my comments.
> 
> * My only real point is that now we have a "recommended" constructor, we should present this first, likewise with the simple example for it, so could you move everything for this up to the top? Then most users will only have to read the first part of this page. After we can talk about general constructors, and all the details for the constraints etc.

Good idea -- I definitely agree that it'll be best for users if we put all the info pertaining to the "recommended" constructor up front.

tsj5:
Current CI failures are fixed by issue #175 / PR #176.

odunbar:
@tsj5 Much prefer the structure now! LGTM - merge away!

PS - you have been extraordinarily unlucky with the CI failures on your PRs :(

PR #162, opened 2022-05-18
Add UQ parameter reading / writing functionality

bielim:
This code adds the module `UQParameters.jl`, which provides functionality for reading UQ parameters from `.toml` files and writing them back to `.toml` files. It also allows for convenient writing of EKP parameter ensembles. The code used to live in `CLIMAParameters` but was moved here due to its dependency on `EnsembleKalmanProcesses.jl`. 

The discussion on whether or not the use of `eval` should be permitted has been postponed to a potential future PR. `eval` is currently not used in the code due to the security leak it introduces, but its use would make the code in its current form as well as its maintenance in the future considerably simpler.


src/file_parsing_uq.jl
```diff
@@ -0,0 +1,507 @@
+using TOML
+using Distributions
+using EnsembleKalmanProcesses.ParameterDistributions
+
+
+# "Public" functions (i.e., functions intended to be called by user):
```
odunbar, line 6: I agree we should add this into a module.

odunbar (changes requested):
Having now gone through it now, there seems little to do here. as the CP dependence doesn't actually get used in the code.

src/file_parsing_uq.jl
```diff
@@ -0,0 +1,507 @@
...
+                      format of the parameter information.
+
+Returns a nested dictionary whose keys are the parameter names (= headers of
+the toml tables) and whose values are dictionaries containing the corresponding
+parameter information (e.g., "prior", "constraint", "value", etc.)
+"""
+function read_parameters(path_to_toml_file::AbstractString)
+    param_dict = TOML.parsefile(path_to_toml_file)
+    return param_dict
+end
```
charleskawczynski, line 32: I recommend just inlining, and delete, this method. using `TOML.parsefile(path_to_toml_file)` is self explanatory, and it's a stable interface

src/file_parsing_uq.jl
```diff
@@ -0,0 +1,507 @@
...
+
+
+"""
+get_parameter_distribution(param_dict, name)
```
charleskawczynski, line 36:
```suggestion
    get_parameter_distribution(param_dict, name)
```

src/file_parsing_uq.jl
```diff
@@ -0,0 +1,507 @@
...
+    # Constructing a parameter distribution requires a prior distribution,
+    # a constraint, and a name.
+    prior = construct_prior(param_dict[name])
+    constraint = construct_constraint(param_dict[name])
+
+    return ParameterDistribution(prior, constraint, name)
+end
+
+"""
+get_parameter_distribution(param_dict, names)
```
charleskawczynski, line 59:
```suggestion
    get_parameter_distribution(param_dict, names)
```

src/file_parsing_uq.jl
```diff
@@ -0,0 +1,507 @@
...
+
+    param_dist_arr = Array{ParameterDistribution}(undef, length(names))
+
+    for (i, name) in enumerate(names)
+        param_dist_arr[i] = get_parameter_distribution(param_dict, name)
+    end
+
+    return combine_distributions(param_dist_arr)
+
+end
```
charleskawczynski, line 81:
```suggestion
function get_parameter_distribution(param_dict::Dict, names::AbstractVector{String})
    param_dist_arr = map(names) do name
        get_parameter_distribution(param_dict, name)
    end
    return combine_distributions(param_dist_arr)
end
```

src/file_parsing_uq.jl
```diff
@@ -0,0 +1,507 @@
...
+
+
+"""
+construct_constraint(param_info)
```
charleskawczynski, line 85:
```suggestion
    construct_constraint(param_info)
```

src/file_parsing_uq.jl
bielim: Neat :)

src/file_parsing_uq.jl
charleskawczynski: FWIW, this function was removed from CLIMAParameters

src/file_parsing_uq.jl
```diff
@@ -0,0 +1,502 @@
...
+    if d.args[1] == Symbol("VectorOfParameterized")
+        # There are multiple distributions described by a
+        # `VectorOfParameterized`
+        return get_vector_of_parameterized(d)
+
+    else
+        # Single distribution
+        return get_distribution_from_expr(d)
+    end
+end
```
charleskawczynski, line 136:
```suggestion
function construct_prior(param_info::Dict)
    @assert(haskey(param_info, "prior"))
    return eval(Meta.parse(param_info["prior"]))
end
```
Can we use this instead?

Project.toml
```diff
@@ -4,6 +4,7 @@ authors = ["CLIMA contributors <clima-software@caltech.edu>"]
 version = "0.8.0"
 
 [deps]
+CLIMAParameters = "6eacf6c3-8458-43b9-ae03-caf5306d3d53"
```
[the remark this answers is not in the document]
bielim: Yes, it has been removed!

src/file_parsing_uq.jl
bielim: Fixed

src/file_parsing_uq.jl
bielim: `file_parsing_uq.jl` had been turned into a module `UQParameters.jl`.

src/file_parsing_uq.jl
bielim: Okay, then I will delete the function here too. If e.g. checks/assertions will become necessary at some point we can also do them after reading the parameters in (e.g, placing them in the functions that require the assertions to pass).

bielim: `read_parameters()` has been deleted.

src/file_parsing_uq.jl
```diff
@@ -0,0 +1,502 @@
...
+        # Constraints are given as a `repeat` expression defining a vector of
+        # constraints of the same kind
+        constr = collect_from_expr(c.args[2], "c", repeat=true)
+        n_constr = c.args[3] # number of repeated constraints
+        return repeat([constr], n_constr)
+
+    else 
+        return collect_from_expr(c, "c")
+    end
+end
```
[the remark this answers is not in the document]
bielim: Closing this after discussion with @charleskawczynski --- we'll leave the code as is for now and might change that later.

src/file_parsing_uq.jl
bielim: Closing this after discussion with @charleskawczynski --- we'll leave the code as is for now and might change it later, once we have agreed on whether using `eval` is permissible.

PR #163, opened 2022-05-25
add N_ens input to Unscented, it can be N+2 or 2N+1

Zhengyu-Huang:
In this PR, we add an argument for Unscented struct and its constructor

The number of sigma points (ensemble size) N_ens can be  = 2N+1 or N+2

In its constructor, the default value is 2N+1

### DIFF
--- a/src/UnscentedKalmanInversion.jl
+++ b/src/UnscentedKalmanInversion.jl
@@ -18,6 +18,7 @@ $(TYPEDFIELDS)
         update_freq::IT = 1,
         modified_unscented_transform::Bool = true,
         prior_mean::Union{AbstractVector{FT}, Nothing} = nothing,
+        sigma_points::String = symmetric
     ) where {FT <: AbstractFloat, IT <: Int}
 
 Construct an Unscented Inversion Process.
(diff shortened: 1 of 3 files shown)


src/UnscentedKalmanInversion.jl
```diff
@@ -18,6 +18,7 @@ $(TYPEDFIELDS)
         update_freq::IT = 1,
         modified_unscented_transform::Bool = true,
         prior_mean::Union{AbstractVector{FT}, Nothing} = nothing,
+        N_ens::IT = 2size(u0_mean, 1)+1
```
ilopezgp, line 21: I am concerned about this being too general an input (Integer). Can we instead add a keyword argument `sigma_points` that is either `symmetric` for 2p+1 or `simplex` for p+2?

src/UnscentedKalmanInversion.jl
```diff
@@ -73,30 +77,62 @@ function Unscented(
...
+        c_weights[1:N_par] .= sqrt(N_par + λ)
+        mean_weights[1] = λ / (N_par + λ)
+        mean_weights[2:N_ens] .= 1 / (2 * (N_par + λ))
+        cov_weights[1] = λ / (N_par + λ) + 1 - α^2 + 2.0
+        cov_weights[2:N_ens] .= 1 / (2 * (N_par + λ))
+
+        if modified_unscented_transform
+            mean_weights[1] = 1.0
+            mean_weights[2:N_ens] .= 0.0
+        end
```
ilopezgp, line 106: I think this is shared between the two options, so we can take it out of the if/else statement.

Zhengyu-Huang: Thanks for the review! I have moved this outside the loop and add sigma_points::String, and update the tests

ilopezgp (approved):
LGTM, thanks @Zhengyu-Huang !! Can you squash commits before merging? [Here](https://clima.github.io/EnsembleKalmanProcesses.jl/dev/contributing/#Squash-and-rebase) is how to do it just in case.

PR #166, opened 2022-06-08
Bugfixes to Documenter config, docs links

tsj5:
This PR fixes two separate bugs which raise warnings when building documentation, as described in issue #167. Thanks to @charleskawczynski for helping me debug this issue!

1) The first bug is that relative paths are not supported as targets of Documenter `@ref` links (Documenter issue [509](https://github.com/JuliaDocs/Documenter.jl/issues/509) -- I *think* the link is generated correctly but doesn't pass the linkcheck stage, but don't quote me on that.)
The fix implemented here is the same as [what was used](https://github.com/JuliaLang/julia/pull/22201) for the julialang docs: the link targets are explicitly marked with an [`@id` tag](https://juliadocs.github.io/Documenter.jl/stable/man/syntax/#Duplicate-Headers), and the `@ref` tag is changed to refer to that `@id` tag, rather than the file's path. 
Ironcially, this could be better documented in the Documenter docs themselves...

2) The second bug only takes place when building docs locally, and results in missing file warnings for files that exist. This is due to Literate.jl, which generates non-functional URLs (containing the string [`<unkown>`](https://github.com/fredrikekre/Literate.jl/blob/cec2506d22f1367bf03fc3487f14009eff2fd03b/src/Literate.jl#L204-L206)) if the config variable [`repo_root_url`](https://fredrikekre.github.io/Literate.jl/v2/outputformats/#Configuration) is not set. Literate.jl's default configuration picks this up correctly when run under common CI services (eg. GH Actions), but not when built locally. 
The fix implemented here is to manually set this path in make.jl only when being run outside of CI (which we test for by testing the existence of a shell environment variable named `CI`).

With the two fixes above, documentation builds locally without raising any warnings, and manually checking confirms the links in question are generated correctly.


odunbar:
```
┌ Warning: invalid local link: unresolved path in ensemble_kalman_inversion.md
│   link.text =
│    1-element Vector{Any}:
│     "Prior distributions"
│   link.url = "parameter_distributions.html"
└ @ Documenter.Writers.HTMLWriter ~/.julia/packages/Documenter/MLty7/src/Writers/HTMLWriter.jl:2081
```
In the latest GH run i still see this warning.

tsj5:
bors try

tsj5:
> ```
> ┌ Warning: invalid local link: unresolved path in ensemble_kalman_inversion.md
> │   link.text =
> │    1-element Vector{Any}:
> │     "Prior distributions"
> │   link.url = "parameter_distributions.html"
> └ @ Documenter.Writers.HTMLWriter ~/.julia/packages/Documenter/MLty7/src/Writers/HTMLWriter.jl:2081
> ```
> 
> In the latest GH run i still see this warning. (and the other invalid local links)

I was misunderstanding the problem in the earlier version of this PR -- this should be completely fixed.

odunbar:
LGTM! Just need to merge from `main`.

odunbar (approved):

PR #176, opened 2022-06-28
Include expected warning text in @test_logs

### LINKED ISSUE #175: Make use of @test_logs less brittle
This issue is raised to describe problems encountered [here](https://github.com/CliMA/EnsembleKalmanProcesses.jl/runs/7084719339?check_suite_focus=true#step:20:416) with getting the test suite to pass for an unrelated PR (#161). I've been able to reproduce it locally on `main` as well (since it originates in a dependency).

### Description

The linked [test suite failure](https://github.com/CliMA/EnsembleKalmanProcesses.jl/runs/7084719339?check_suite_focus=true#step:6:356) arises from a unit test of the form [`@test_logs (:warn,)`](https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/0fa998e10556ab2117c78abd2a3cda5f541c1bb7/test/EnsembleKalmanProcess/runtests.jl#L522): when run under GH actions, the test gets two warnings, instead of the single warning it expected, and fails. 

The second warning isn't related to EKP: it's a deprecation warning raised by MvNormal:
```
┌ Warning: `dim(a::AbstractMatrix)` is deprecated, use `LinearAlgebra.checksquare(a)` instead.
│   caller = MvNormal at mvnormal.jl:186 [inlined]
└ @ Core ~/.julia/packages/Distributions/9Albf/src/multivariate/mvnormal.jl:186
```
I can reproduce this locally: it arises because PDMats.jl deprecated `dim` in its [latest release](https://github.com/JuliaStats/PDMats.jl/releases/tag/v0.11.14), two days ago as of this writing, and Distributions.jl hasn't been updated.

### Proposed fix

I propose to fix this by modifying EKP's usage of `@test_logs` to 1) test for the text of the specific, expected warning message, and 2) to ignore any other warnings via `match_mode=:any`. 

- It's not feasible to pin the version of PDMats in EKP's Project.toml; it's a dependency of a dependency (LinearAlgebra) and we'd have to remember to unpin and update it when Distributions.jl fixes its deprecation warnings. 
- At the same time, we should keep `--depwarn=yes` in EKP's test suite, to warn about deprecations in code we control.
- `@test_logs` provides some functionality for [filtering](https://docs.julialang.org/en/v1/stdlib/Test/#Test.@test_logs) log messages, but the filters are applied in one-to-one order with the log messages raised; there doesn't appear to be a way to filter out logs that may or may not be present, except through the use of `match_mode=:any`.

tsj5:
# PULL REQUEST

### Purpose
Change use of `@test_logs` to allow unit tests to pass when dependencies raise deprecation warnings. The writeup in issue #175 makes the case that the solution implemented here is the best way to deal with the case where the deprecation warning is raised in code we don't control. 

### Content
For all uses of `@test_logs` in unit tests,
1) add a filter to match the text of the expected warning message,
2) add `match_mode=:any`, to allow the test to pass when deprecation errors are raised.

### Benefits
Unit test suite now passes without needing to wait for the deprecations in Distributions.jl to be fixed in a new release;

### Risks
If other, "serious" warnings are raised in addition to the warning being tested for, this implementation will give a false negative. As discussed in issue #175, as far as I know there's no way to optionally filter out deprecation warnings only.

## Linked Issues
- Closes #175

## PR Checklist
- [X] This PR has a corresponding issue OR is linked to an SDI.
- [X] I have followed CliMA's codebase [contribution](https://clima.github.io/ClimateMachine.jl/latest/Contributing/) and [style](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) guidelines OR N/A.
- [X] I have followed CliMA's [documentation policy](https://github.com/CliMA/policies/wiki/Documentation-Policy).
- [X] I have checked all issues and PRs and I certify that this PR does not duplicate an open PR.
- [X] I linted my code on my local machine prior to submission OR N/A.
- [X] Unit tests are included OR N/A.
- [X] Code used in an integration test OR N/A.
- [X] All tests ran successfully on my local machine OR N/A.
- [X] All classes, modules, and function contain docstrings OR N/A.
- [X] Documentation has been added/updated OR N/A.

### DIFF
--- a/test/EnsembleKalmanProcess/runtests.jl
+++ b/test/EnsembleKalmanProcess/runtests.jl
@@ -516,10 +516,13 @@ const EKP = EnsembleKalmanProcesses
         for i in 2:7
             g[:, i] .= NaN
         end
-        @test_logs (:warn,) split_indices_by_success(g)
+        @test_logs (:warn, r"More than 50% of runs produced NaNs") match_mode = :any split_indices_by_success(g)
 
         u = rand(10, 4)
-        @test_logs (:warn,) sample_empirical_gaussian(u, 2)
+        @test_logs (:warn, r"Sample covariance matrix over ensemble is singular.") match_mode = :any sample_empirical_gaussian(
+            u,
+            2,
+        )
         @test_throws PosDefException sample_empirical_gaussian(u, 2, inflation = 0.0)
     end
 


ilopezgp (approved):
LGTM

PR #180, opened 2022-07-08
Add cutoff to SEC localizer

ilopezgp:
# PULL REQUEST

## Purpose and Content

Adds the option to apply a cutoff below which SEC-corrected covariances are set to 0, sparsifying the covariance matrix. This follows the logic of kernels used in data assimilation which have a compact support, like the Gaspari-Cohn function.

## Benefits and Risks

Explore the usefulness of correlation sparseness, addition of a novel localization function (not used yet in the literature?). So far it is looking good! (See plot below).

## PR Checklist
- [x] I linted my code on my local machine prior to submission OR N/A.
- [x] Unit tests are included OR N/A.
- [x] Code used in an integration test OR N/A.
- [x] All tests ran successfully on my local machine OR N/A.
- [x] All classes, modules, and function contain docstrings OR N/A.
- [ ] Documentation has been added/updated OR N/A.

### DIFF
--- a/examples/Localization/localization_example_lorenz96.jl
+++ b/examples/Localization/localization_example_lorenz96.jl
@@ -107,6 +107,15 @@ for i in 1:N_iter
     EKP.update_ensemble!(ekiobj_sec, g_ens, deterministic_forward_map = true)
 end
 
+# Test SEC with cutoff
+ekiobj_sec_cutoff =
+    EKP.EnsembleKalmanProcess(initial_ensemble, y, Γ, Inversion(); rng = rng, localization_method = SEC(1.0, 0.1))
+
+for i in 1:N_iter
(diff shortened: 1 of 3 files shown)


ilopezgp:
result

@eviatarbach :)

eviatarbach:
This is great! Just one comment: in the comparison figure, can you use the same value of alpha for SEC with/without cutoff?

ilopezgp:
> This is great! Just one comment: in the comparison figure, can you use the same value of alpha for SEC with/without cutoff?

I think I am, I am using `SEC(1.0, 0.1)` and `SEC(1.0)`. The first parameter is alpha, the second is the cutoff.

eviatarbach (approved):

PR #186, opened 2022-08-22
Unlink real inputs in ParameterDistributions.

### LINKED ISSUE #185: Unlink parameter types
https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/e30f2f959b7894ba6207e629742f0f9db505f0e5/src/ParameterDistributions.jl#L750-L753

Simple cases can break when e.g., the four inputs are `0,1,-Inf,Inf` which is viewed as `Int64, Int64, Float64, Float64` (but above requires all arguments are the same type `::FT`).

## Possible solution 
Replace with 
```julia
mu_c::Real, 
sigma_c::Real,
lower_bound::Real,
upper_bound::Real
```

Note - this could be more pervasive than just the example I found

ilopezgp:
# PULL REQUEST

## Purpose and Content

Unlinks the type of real inputs to ParameterDistributions functions, allowing the use of `Int`s and infinites.

## Linked Issues
- Closes #185

### DIFF
--- a/src/EnsembleKalmanInversion.jl
+++ b/src/EnsembleKalmanInversion.jl
@@ -37,7 +37,7 @@ end
     find_ekp_stepsize(
         ekp::EnsembleKalmanProcess{FT, IT, Inversion},
         g::AbstractMatrix{FT};
-        cov_threshold::FT = 0.01,
+        cov_threshold::Real = 0.01,
     ) where {FT, IT}
 
 Find largest stepsize for the EK solver that leads to a reduction of the determinant of the sample
@@ -46,14 +46,13 @@ covariance matrix no greater than cov_threshold.
 function find_ekp_stepsize(
     ekp::EnsembleKalmanProcess{FT, IT, Inversion},
     g::AbstractMatrix{FT};
-    cov_threshold::FT = 0.01,
+    cov_threshold::Real = 0.01,
 ) where {FT, IT}
+    @assert cov_threshold > -eps(FT) "The limiting covariance reduction threshold cannot be negative."
     accept_stepsize = false
-    if !isempty(ekp.Δt)
-        Δt = deepcopy(ekp.Δt[end])
-    else
-        Δt = FT(1)
-    end
+
+    Δt = !isempty(ekp.Δt) ? deepcopy(ekp.Δt[end]) : FT(1)
+
     # final_params [N_par × N_ens]
     cov_init = cov(get_u_final(ekp), dims = 2)
     while accept_stepsize == false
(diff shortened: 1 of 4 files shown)


ilopezgp:
Would this work @odunbar ?

src/EnsembleKalmanInversion.jl
```diff
@@ -46,14 +46,13 @@ covariance matrix no greater than cov_threshold.
 function find_ekp_stepsize(
     ekp::EnsembleKalmanProcess{FT, IT, Inversion},
     g::AbstractMatrix{FT};
-    cov_threshold::FT = 0.01,
+    cov_threshold::Real = 0.01,
 ) where {FT, IT}
+    @assert cov_threshold > -eps(FT) "The limiting covariance reduction threshold cannot be negative."
```
odunbar, line 51: `eps(Real)`?

src/EnsembleKalmanInversion.jl
```diff
@@ -77,8 +76,8 @@ end
         u::AbstractMatrix{FT},
         g::AbstractMatrix{FT},
         y::AbstractMatrix{FT},
-        obs_noise_cov::Union{AbstractMatrix{FT}, UniformScaling{FT}},
-    ) where {FT <: Real, IT}
+        obs_noise_cov::Union{AbstractMatrix{CT}, UniformScaling{CT}},
```
odunbar, line 79: I think maybe `obs_noise_cov` should stay linked? Was there a case you had in mind for when they differ?

Just looking ahead we do `(cov_gg + obs_noise_cov) \ (y - g)` and presumably we don't want any `FT` vs `CT` type issues here.

odunbar (commented):
LGTM - and 1 small comment

This is likely my fault for initiating the convention of `(...::FT...) where FT` while forgetting the restrictions it imposes

src/EnsembleKalmanInversion.jl
ilopezgp: `eps(Real)` does not exist, unfortunately, since it does not specify the arithmetic precision.

src/EnsembleKalmanInversion.jl
ilopezgp: I was thinking of this case:

```
julia> using Random

julia> using LinearAlgebra

julia> cov1 = I
UniformScaling{Bool}
true*I

julia> cov2 = 1.0 * I
UniformScaling{Float64}
1.0*I

julia> y = rand(5)
5-element Vector{Float64}:
 0.8876949829922091
 0.9351539077673359
 0.992727497437651
 0.6203034498205577
 0.9959225311210873

julia> g = rand(5)
5-element Vector{Float64}:
 0.9247820413465523
 0.6061500089264417
 0.739520036070634
 0.3345260433372921
 0.606807569247004

julia> res1 = cov1 \ (y - g)
5-element Vector{Float64}:
 -0.03708705835434323
  0.3290038988408942
  0.253207461367017
  0.28577740648326566
  0.38911496187408334

julia> res2 = cov2 \ (y - g)
5-element Vector{Float64}:
 -0.03708705835434323
  0.3290038988408942
  0.253207461367017
  0.28577740648326566
  0.38911496187408334

julia> res1 == res2
true
```

What I can do is cast the result to FT to ensure type stable computation, but I prefer to not be so strict about input typing.

PR #196, opened 2022-08-26
ParameterDistributions.jl improvements

### LINKED ISSUE #184: Overloading == for ParameterDistributions
```julia
unif1 = ParameterDistribution(
                      Dict(
                          "distribution" => Parameterized(Uniform(0,2*π)),
                          "constraint" => no_constraint(),
                          "name" => "uniform",
                      )
                  )
unif2 = ParameterDistribution(
                      Dict(
                          "distribution" => Parameterized(Uniform(0,2*π)),
                          "constraint" => no_constraint(),
                          "name" => "uniform",
                      )
                  )
```
satisfies
```julia
julia> get_distribution(unif) == get_distribution(unif2) 
true
julia> get_all_constraints(unif) == get_all_constraints(unif2) 
true
julia> get_name(unif) == get_name(unif2)
true
```
but 
```julia
julia> unif == unif2 
false
```
this is somewhat annoying for various tests/checks

## Solution?
As `==` defaults to `===` without overloading, the behaviour will give `false`. So a solution seems to be to overload `==` to check all the internal components, e.g.,
```julia
Base.:(==)(lhs ::ParameterDistribution, rhs ::ParameterDistribution) = 
(lhs.distribution == rhs.distribution) && 
(lhs.constraint == rhs.constraint) &&
(lhs.name == rhs.name)
```

### LINKED ISSUE #192: constrained_gaussian type constructor for marginal-defined multivariate distributions
## Purpose
To extend `constrained_gaussian` constructor to multivariate distributions, having the same constraints in every dimension. It's useful for helping to define multidimensional problems defined from marginals that arise quite often

## Currently
we have 
```julia
constrained_gaussian("name", mean, var, lower_bound, upper_bound)
```
Which finds `comp_mean`, `comp_var` based on the constraints implied by the bounds
```julia 
ParameterDistribution(
    Dict(
        "distribution" => Parameterized(Normal(comp_mean, comp_var)),
        "constraint" => <implied by bounds>,
        "name" => "name",
    )
)
```
## Extension
Called by 
```julia
constrained_gaussian("name", mean, var, lower_bound, upper_bound, repeat=12)
```
That would produce
```julia
ParameterDistribution(
    Dict(
        "distribution" => VectorOfParameterized(repeat([Normal(comp_mean, comp_var)],12)),
        "constraint" => repeat([<implied by bounds>],12)
        "name" => "name",
    )
)
```

odunbar:
## Purpose 
Fixes #184 and #192, also small update to remove warning in DataContainers/ParameterDistributions test. This works towards a better interface for ParameterDistributions. 

## Content
- Extending `constrained_gaussian` to multidimensional distirbutions by adding a `repeats=N` keyword with `N=1` default.
- Adds Types constraints into `NoConstraint` `BoundedBelow` `BoundedAbove` and `Bounded`
- Adds "parameters" to constraints, containing a Dict (or nothing). allows extraction of values of interest. for example 
`x = bounded_below(3.0)`  stores  `x.parameters = Dict("lower_bound" => 3.0)`
- extending `==` to Constraints (based on the above `parameters` and `Types`, as constraints are functions so hard to equate) 
- extending `==` to ParameterDistributions for easier testing calls
- Miscellaneous: Removed the `MvNormal(int,real)` calls that are now deprecated, replacing with `MvN(zeros(int), real*I)`

### DIFF
--- a/src/ParameterDistributions.jl
+++ b/src/ParameterDistributions.jl
@@ -5,6 +5,7 @@ using Distributions
 using Statistics
 using Random
 using Optim, QuadGK
+using DocStringExtensions
 
 #import (to add definitions)
 import StatsBase: mean, var, cov, sample
@@ -18,17 +19,19 @@ export ConstraintType
 #objects
 export Parameterized, Samples, VectorOfParameterized
 export ParameterDistribution
-export Constraint
+export Constraint, NoConstraint, BoundedBelow, BoundedAbove, Bounded
 
 #functions
 export get_name, get_distribution, ndims, get_dimensions, get_all_constraints, get_n_samples
 export sample
 export no_constraint, bounded_below, bounded_above, bounded
+export get_bounds, get_constraint_type
 export transform_constrained_to_unconstrained, transform_unconstrained_to_constrained
 export get_logpdf, batch
 
 export combine_distributions
 export constrained_gaussian
+
 ## Objects
 # for the Distribution
 abstract type ParameterDistributionType end
@@ -41,6 +44,7 @@ A distribution constructed from a parametrized formula (e.g Julia Distributions.
 struct Parameterized <: ParameterDistributionType
     distribution::Distribution
 end
+
 
 """
     Samples{FT <: Real} <: ParameterDistributionType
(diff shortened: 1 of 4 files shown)


src/ParameterDistributions.jl
```diff
@@ -69,19 +72,23 @@ end
...
     Constraint <: ConstraintType
 
 Class describing a 1D bijection between constrained and unconstrained spaces.
 """
-struct Constraint <: ConstraintType
+struct Constraint{T} <: ConstraintType
     constrained_to_unconstrained::Function
     c_to_u_jacobian::Function
     unconstrained_to_constrained::Function
+    parameters::Union{Dict, Nothing}
```
ilopezgp, line 89: Can you add docstrings on top of each element of this struct for the API docs?

Also, I think we might be overusing the word `parameters` here, since these are the `parameters defining a parameter transformation`. Could we use another term, perhaps `bounds`?

src/ParameterDistributions.jl
```diff
@@ -154,11 +163,16 @@ function bounded(lower_bound::FT, upper_bound::FT) where {FT <: Real}
...
             jacobian = (x -> 1.0 / (upper_bound - x) + 1.0 / (x - lower_bound))
             u_to_c = (x -> upper_bound - (upper_bound - lower_bound) / (exp(x) + 1))
-            return Constraint(c_to_u, jacobian, u_to_c)
+            parameters = Dict("lower_bound" => lower_bound, "upper_bound" => upper_bound)
+            return Constraint{Bounded}(c_to_u, jacobian, u_to_c, parameters)
         end
     end
 end
 
+get_parameters(c::CType) where {CType <: ConstraintType} = c.parameters
```
ilopezgp, line 172: similar here, maybe rename to `get_bounds`.

src/ParameterDistributions.jl
odunbar: It is any parameter that could be used to build a constraint - currently they are bounds but it could also be e.g. a scaling parameter in the sigmoid. So may not only be bounds. I can definitely rename though

odunbar: I think I'm happy to use `bounds` and change the term if we find a good use-case for when the parameter value is not a bound

src/ParameterDistributions.jl
odunbar: done

ilopezgp (approved):
LGTM

Issue #24, opened 2021-04-22
Returning constrained parameters


odunbar:
Our current getter methods for EKP provide exactly the stored computational parameters. However in typical use with the G model requiring constrained parameters, it would be nice to have a getter function for this (otherwise one needs to apply it separately)

i.e for some `ekp::EnsembleKalmanProcess` object, and a created `prior::ParameterDistribution`  currently one applies
```julia 
computational_params = get_u_final(ekp)
physical_params = transform_unconstrained_to_constrained(prior,computational_params)
model_outputs = run_model_ensemble(physical_params)
```
## Solution options
We wish to apply the transformation internally. 

We could add new getters `get_u_constrained` and `get_u_constrained_final`. 

Or we could have a flag to `return_constrained` in the current getter

ilopezgp:
Should it be

`physical_params = transform_unconstrained_to_constrained(prior,computational_params)` instead?

odunbar:
corrected!

odunbar:
Note also, when creating the `create_initial_ensemble` this appears to also be done in the unconstrained space.

PR #208, opened 2022-09-20
Add more useful getters, improve API docs.

ilopezgp:
# PULL REQUEST

## Purpose and Content

- Improves some imprecise/too generic API docs,
- Adds more useful getter functionality.


odunbar (commented):
Thanks, I much prefer the descriptions
Two comments on getters. 
1. I think for we should really have getters for the final covariance too. Particularly for UKI it is really buried inside the process as `ukiobj.process.uu_cov[end]`. Suggestion: `get_u_cov`, `get_u_cov_final` (for EKI/EKS this is just the empirical cov for `dims=2`)
2. I've also been for a long time, wanting something to return the transformed ensemble, as users often forget the transform. Suggested notation would be `get_\phi_final` `get_\phi` `get_\phi_mean` etc. that simply return `transform_unconstrained_to_constrained` applied to `get_u_final`, `get_u`, `get_u_mean` etc. What do you think?

ilopezgp:
> Thanks, I much prefer the descriptions Two comments on getters.
> 
> 1. I think for we should really have getters for the final covariance too. Particularly for UKI it is really buried inside the process as `ukiobj.process.uu_cov[end]`. Suggestion: `get_u_cov`, `get_u_cov_final` (for EKI/EKS this is just the empirical cov for `dims=2`)
> 2. I've also been for a long time, wanting something to return the transformed ensemble, as users often forget the transform. Suggested notation would be `get_\phi_final` `get_\phi` `get_\phi_mean` etc. that simply return `transform_unconstrained_to_constrained` applied to `get_u_final`, `get_u`, `get_u_mean` etc. What do you think?

I agree, those sound like good suggestions. I will add those to this PR.

ilopezgp:
@odunbar This should be ready!

ilopezgp:
Closes #24

src/UnscentedKalmanInversion.jl
```diff
@@ -666,6 +666,14 @@ function get_u_mean_final(uki::EnsembleKalmanProcess{FT, IT, Unscented}) where {
     return uki.process.u_mean[end]
 end
 
+"""
+    get_u_cov_final(uki::EnsembleKalmanProcess{FT, IT, Unscented}) where {FT <: AbstractFloat, IT <: Int}
+
+Returns the unconstrained parameter covariance at the final iteration.
+"""
+function get_u_cov_final(uki::EnsembleKalmanProcess{FT, IT, Unscented}) where {FT <: AbstractFloat, IT <: Int}
```
odunbar, line 674: Instead of writing the method `get_u_cov_final` for UKI, we should instead write the method `get_u_cov` for uki:
```
function get_u_cov(uki::EnsembleKalmanProcess{FT,IT,Unscented}, iteration::IT) where {FT <: AbstractFloat, IT <: Integer}
    return uki.process.uu_cov[it]
end
```
Then we can use the `get_u_cov_final` method from ekp?

odunbar (commented):
I think you need to change the UKI getter method - we get more functionality by only creating the method `get_u_cov` for UKI. 

Other changes are great thank you!

src/UnscentedKalmanInversion.jl
ilopezgp: Sounds good, I will do the same for get_u_mean_final.

odunbar (approved):
LGTM

PR #214, opened 2022-09-28
new tests and constructors

### LINKED ISSUE #212: New constructor for Unscented
https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/d9bfc17ece586f652805698807873009f1d3a5e6/src/UnscentedKalmanInversion.jl#L74-L82

Currently the `Unscented` constructor needs a mean and covariance. Thanks to our new interfaces, the users do not need to know the underlying computational-space Gaussians that are used. I think the following constructor would work 

```julia
function Unscented( 
     prior::ParameterDistribution,
     α_reg::FT = 1.0, 
     update_freq::IT = 1, 
     prior_mean::Union{AbstractVector{FT}, Nothing} = nothing, 
     sigma_points::String = "symmetric", 
 ) where {FT <: AbstractFloat, IT <: Int} 

u0_mean = Vector{FT}(mean(prior)) # mean of computational gaussian
uu0_cov = Matrix{FT}(cov(prior)) # cov of computational gaussian

return Unscented(u0_mean, uu0_cov,...)
end
```
Likewise I'm not-sure the difference in the u0_mean and the prior_mean here

odunbar:
## Purpose 
Unifies the constructors with the new prior interfaces. for examples users will not have to calculate means or covariances in computational space of the prior for to build `Unscented` or `Sampler` objects. instead we pass the `prior`.

## Content 
- new (additional) constructors for `Unscented` and `Sampler` objects
- unit tests

## Linked Issues
- Closes #212

### DIFF
--- a/src/UnscentedKalmanInversion.jl
+++ b/src/UnscentedKalmanInversion.jl
@@ -170,6 +170,17 @@ function Unscented(
     )
 end
 
+function Unscented(prior::ParameterDistribution; kwargs...)
+
+    u0_mean = Vector(mean(prior)) # mean of unconstrained distribution
+    uu0_cov = Matrix(cov(prior)) # cov of unconstrained distribution
+
+    return Unscented(u0_mean, uu0_cov; kwargs...)
+
+end
+
+
+
 # outer constructors
 function EnsembleKalmanProcess(
     obs_mean::AbstractVector{FT},
--- a/src/EnsembleKalmanSampler.jl
+++ b/src/EnsembleKalmanSampler.jl
@@ -20,6 +20,14 @@ struct Sampler{FT <: AbstractFloat} <: Process
     prior_cov::Union{AbstractMatrix{FT}, UniformScaling{FT}}
 end
 
+function Sampler(prior::ParameterDistribution)
+    mean_prior = Vector(mean(prior))
+    cov_prior = Matrix(cov(prior))
+    FT = eltype(mean_prior)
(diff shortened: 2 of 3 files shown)


src/UnscentedKalmanInversion.jl
```diff
@@ -170,6 +170,31 @@ function Unscented(
...
+        u0_mean,
+        uu0_cov;
+        α_reg = α_reg,
+        update_freq = update_freq,
+        modified_unscented_transform = modified_unscented_transform,
+        prior_mean = u0_mean,
+        sigma_points = sigma_points,
+    )
+
+end
```
ilopezgp, line 194:
```suggestion
function Unscented(prior::ParameterDistribution; kwargs...)
    u0_mean = Vector(mean(prior)) # mean of unconstrained distribution
    uu0_cov = Matrix(cov(prior)) # cov of unconstrained distribution
    return Unscented(u0_mean, uu0_cov; kwargs...)
end
```
A bit shorter

odunbar: This almost works, I think we will need to do 
```julia
return Unscented(u0_mean, uu0_cov; prior_mean = u0_mean, kwargs...)
```

odunbar: I'll just do your suggestion for now as there is no benefit of mine really...

ilopezgp (approved):

PR #228, opened 2022-11-03
Adding sinusoid example

### LINKED ISSUE #183: Add a simple pedagogical example with finding the parameters of a sinusoid
As part of @anagha548's SURF project, her, @odunbar, and I have been playing with the simple inverse problem

`f(A, c) = A*sin(t) + c`

We take the observations to be a noisy realization from time 0 to 2*pi, and with a random phase. We take the observables to be the mean of this curve (to find the vertical shift) and the maximum minus the minimum (to find the amplitude).

I think this is a simple pedagogical example which could be helpful in the documentation as a problem to introduce EKI.

eviatarbach:
Closes #183


eviatarbach:
bors try

examples/Sinusoid/sinusoid_example.jl
```diff
@@ -0,0 +1,100 @@
...
+initial_ensemble = EKP.construct_initial_ensemble(rng, prior, N_ensemble)
+
+ensemble_kalman_process = EKP.EnsembleKalmanProcess(initial_ensemble, y, Γ, Inversion(); rng = rng)
+nothing # hide
+
+# We are now ready to carry out the inversion. At each iteration, we get the
+# ensemble from the last iteration, apply ``G(\theta)`` to each ensemble member,
+# and apply the Kalman update to the ensemble.
+for i in 1:N_iterations
+    params_i = get_u_final(ensemble_kalman_process)
```
odunbar, line 75: This is actually a typo, it must be `get_\phi_final(priors, ensemble_kalman_process)` (as you have used a lognormal prior in the prior, so get_u_final will return the unconstrained parameters).

odunbar: I guess it still works because of the problem spec having high mass far from the boundary

odunbar (commented):
Looks like a great start! 

I have made a few stylistic comments but there are a couple of substance too: Mainly there are 3 things to do
- replace `get_u_final` with `get_\phi_final` (see my comment).
- Add sections "setting up the problem" and "applying the algorithm" as a lot of the code here is just setup
- Could you add 2 gifs of this example to the index page? One showing the collapse in the constrained parameter-parameter space, and another showing the model ensembles. 1 frame = 1 iteration. I don't think much setup is required here, just a title saying "learning the amplitude and shift of a sine curve"

Thank you!

docs/src/index.md
```diff
@@ -15,6 +15,11 @@ ParameterDistributions.jl                   | Structures to hold prior and poste
 DataContainers.jl                           | Structure to hold model parameters and outputs
 Localizers.jl                               | Covariance localization kernels
 
+## Learning the amplitude and vertical shift of a sine curve
+![Ensemble of parameter estimates by iteration](assets/sinusoid_example_params.gif)![Ensemble of resulting curves by iteration](assets/sinusoid_example_curves.gif)
```
odunbar, line 19: is there a way of putting these side-by-side? Also, the two plots aren't in sync for me.

odunbar, line 19: Also, do you think it's ok that the truth draws new samples in every frame of the gif? We don't do this in practice, but then again, it is a random function so maybe it illustrates this well. I'll let you pick!

odunbar (commented):
Hey! just a couple more notes:

docs/src/index.md
eviatarbach: Hmm, I am not having the sync problem, and I don't know if there's a way to make them sync. I think the easiest way may be to combine them into a single gif. I will do that.

odunbar (approved):
LGTM! 

I have one single concern left, which is that people will say the ensemble 'misses the truth', without knowing this is due to the bias of a single data sample.

eviatarbach:
Yes, what should we do about this? Mention it? I think reducing the observation error would make it converge even faster, which makes the animation less interesting.

odunbar:
https://github.com/openjournals/joss-reviews/issues/4869

Issue #230, opened 2022-11-08
Implement Stochastic Update


costachris:
Aim: Add functionality to perform stochastic updates (in parameter space) with contraction. Update option should be available across all `EnsembleKalmanProcesses` and take the from:
$$u_{n+1} = a*u_n + xi_{n} \qquad (1)$$
**Stochastic Ensemble Kalman Update**
The stochastic modification changes the prediction step in the following manner (*Huang et al. 2022* eqn. 41)

$$  \widehat{m}\_{n+1} = m_{n} ; \qquad \widehat{u}^{j}\_{n + 1} = \widehat{m}\_{n+1} + \sqrt{\frac{1}{1 - \Delta{\tau}}} \left(u^{j}\_{n} - m_{n} \right) \qquad (2)$$

where $m$ is the ensemble average and $\Delta \tau$ is a hyperparameter lying in the range $(0, 1)$. The RHS of equation 2 is equivalent to adding random noise directly, and equates to $xi_{n}$ in eqn. 1. The forward map $G$ must be computed after the prediction step (eqn. 2) before the analysis step (updating covariance estimate Cov(G, G), Cov(u, G), etc.)

costachris:
**Proposed Solution:**
Create a common `stochastic_eki_update` function that can be employed by all `EnsembleKalmanProcesses`, which modifies the existing updates by adding a step equivalent to eqn. 2 above. The overarching `update_ensemble!` method for each `EnsembleKalmanProcesses` will then take a flag (default to `false`) as to whether to apply a stochastic update. The first iteration will apply eqn. 2 to the initial parameter sampling from  `construct_initial_ensemble` (since G hasn't been evaluated yet), and subsequent iterations will apply both eqn. 2 and the analysis step (updating covariances, etc.).

Zhengyu-Huang:
@costachris  Thank you for the great effort！ Yes, in my implementation, $\Delta \tau = \Delta t$ , and I need it between 0 and 1, I generally use 0.5.  If the stochastic_update is on, set dt = min(0.5, dt)   

$xi_n \sim N(0, \frac{\Delta \tau }{1 - \Delta \tau} C_n)$, then both implementations are the same, you can also choose $xi_n \sim N(0, \frac{\Delta \tau }{1 - \Delta \tau} C_0)$

odunbar:
Yeah looks good, Implementing this within each process `update_ensemble!` would be fine.

[...]

PR #233, opened 2022-11-16
Inflation

costachris:
## Purpose 
This PR adds the functionality of performing stochastic parameter updates in the form of multiplicative and additive inflation.

## To-do
- [x] Finalize stochastic update for EKI. 
- [x] Port stochastic update to UKI, Sampler , and Sparse EKI.
- [x] ~Make stochastic update the default~.
- [x] Update tests and check examples/. 
- [x] Add documentation

### DIFF
--- a/docs/src/inflation.md
+++ b/docs/src/inflation.md
@@ -1,0 +1,60 @@
+# Inflation 
+Inflation is an approach that slows down collapse in ensemble Kalman methods.
+Two distinct forms of inflation are implemented in this package. Both involve perturbing the ensemble members following the standard update rule of the chosen Kalman process.
+Multiplicative inflation expands ensemble members away from their mean in a
+deterministic manner, whereas additive inflation hinges on the addition of stochastic noise to ensemble members.
+
+For both implementations, a scaling factor ``s`` is included to extend functionality to cases with mini-batching. 
+The scaling factor ``s`` multiplies the artificial time step ``\Delta t`` in the inflation equations to account for sampling error. For mini-batching, the scaling factor should be:
+```math
+    s = \frac{|B|}{|C|}
+```
+where `` |B| `` is the mini-batch size and `` |C| `` is the full dataset size.
+
+## Multiplicative Inflation 
+Multiplicative inflation effectively scales parameter vectors in parameter space, such that the perturbed
+ensemble remains in the linear span of the original ensemble. The implemented update equation follows
+[Huang et al, 2022](https://arxiv.org/abs/2204.04386) eqn. 41:
+
+```math
+\begin{aligned}
+    m_{n+1} = m_{n} ; \qquad u^{j}_{n + 1} = m_{n+1} + \sqrt{\frac{1}{1 - s \Delta{t}}} \left(u^{j}_{n} - m_{n} \right) \qquad (1)
+\end{aligned}
+```
+where ``m`` is the ensemble average. In this way,
+the parameter covariance is inflated by a factor of ``\frac{1}{1 - s \Delta{t}}``, while the ensemble mean remains fixed.
+```math
+     C_{n + 1} = \frac{1}{1 - s \Delta{t}} C_{n} \qquad (2)
+```
+
+Multiplicative inflation can be used by flagging the `update_ensemble!` method as follows:
+```julia
+    EKP.update_ensemble!(ekiobj, g_ens; multiplicative_inflation = true, s = 1.0)
+```
+
+## Additive Inflation 
+Additive inflation is implemented by systematically adding stochastic perturbations to the parameter ensemble in the form of Gaussian noise. Additive inflation breaks the linear subspace property, meaning the parameter ensemble can evolve outside of the span of the initial ensemble. In additive inflation, the ensemble is perturbed in the following manner after the standard Kalman update:
+
+```math
+     u_{n+1} = u_n + \zeta_{n} \qquad (3) \\
+    \zeta_{n} \sim N(0, \frac{s \Delta{t} }{1 - s \Delta{t}} C_n) \qquad (4)
+```
+This inflates the parameter covariance by a factor of ``\frac{1}{1 - s \Delta{t}}`` as in eqn. 2 , while the ensemble mean remains fixed.
+
+Additive inflation can be used by flagging the `update_ensemble!` method as follows:
+```julia
+    EKP.update_ensemble!(ekiobj, g_ens; additive_inflation = true, s = 1.0)
+```
+Alternatively, the prior covariance matrix may be used to generate additive noise, following:
+```math
+    \zeta_{n} \sim N(0, \frac{s \Delta{t} }{1 - s \Delta{t}} C_{0}) \qquad (5)
+```
+This results in an additive increase in the parameter covariance by `` \frac{s \Delta{t} }{1 - s \Delta{t}} * C_{0}`` , while the mean remains fixed.
+```math
+     C_{n + 1} = C_{n} + \frac{s \Delta{t} }{1 - s \Delta{t}} C_{0} \qquad (6)
+```
+
+Additive inflation using the scaled prior covariance (parameter covariance of initial ensemble) can be used by flagging the `update_ensemble!` method as follows:
+```julia
+    EKP.update_ensemble!(ekiobj, g_ens; additive_inflation = true, use_prior_cov = true, s = 1.0)
+```
--- a/src/EnsembleKalmanProcess.jl
+++ b/src/EnsembleKalmanProcess.jl
@@ -13,7 +13,7 @@ export get_u, get_g, get_ϕ
 export get_u_prior, get_u_final, get_g_final, get_ϕ_final
 export get_N_iterations, get_error, get_cov_blocks
 export get_u_mean, get_u_cov, get_g_mean, get_ϕ_mean
-export get_u_mean_final, get_u_cov_final, get_g_mean_final, get_ϕ_mean_final
+export get_u_mean_final, get_u_cov_prior, get_u_cov_final, get_g_mean_final, get_ϕ_mean_final
 export compute_error!
 export update_ensemble!
 export sample_empirical_gaussian, split_indices_by_success
@@ -239,6 +239,15 @@ function get_u_cov(ekp::EnsembleKalmanProcess, iteration::IT) where {IT <: Integ
 end
 
 """
+    get_u_cov_prior(ekp::EnsembleKalmanProcess)
+
+Returns the unconstrained parameter sample covariance for the initial ensemble.
+"""
+function get_u_cov_prior(ekp::EnsembleKalmanProcess)
+    return cov(get_u_prior(ekp), get_u_prior(ekp), dims = 2)
+end
+
+"""
     get_g_mean(ekp::EnsembleKalmanProcess, iteration::IT) where {IT <: Integer}
 
 Returns the mean forward map evaluation at the given iteration.
@@ -300,7 +309,7 @@ get_ϕ_mean_final(prior::ParameterDistribution, ekp::EnsembleKalmanProcess) =
 """
     get_u_cov_final(ekp::EnsembleKalmanProcess)
 
-Get the mean unconstrained parameter at the last iteration.
+Get the mean unconstrained parameter covariance at the last iteration.
 """
 get_u_cov_final(ekp::EnsembleKalmanProcess) = get_u_cov(ekp, size(ekp.u, 1))
 
@@ -327,6 +336,15 @@ Return process type of EnsembleKalmanProcess.
 function get_process(ekp::EnsembleKalmanProcess)
     return ekp.process
 end
+
+"""
+    get_localizer(ekp::EnsembleKalmanProcess)
+Return localizer type of EnsembleKalmanProcess.
+"""
+function get_localizer(ekp::EnsembleKalmanProcess)
+    return Localizers.get_localizer(ekp.localizer)
+end
+
 
 """
     construct_initial_ensemble(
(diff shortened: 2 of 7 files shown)


Zhengyu-Huang (commented):
Thanks for your great work!

Is "stochastic_scaling_factor"  N_batch/N ?

Could you also add your test in the PR?

costachris:
> Thanks for your great work!
> 
> Is "stochastic_scaling_factor" N_batch/N ?
> 
> Could you also add your test in the PR?

Yes, ideally the user would set `stochastic_scaling_factor` = ` N_batch/N` when calling the API, since ie. N_batch and dataset size aren't attributes of EKP objects, so they have to be provided externally.

src/EnsembleKalmanInversion.jl
```diff
@@ -198,3 +198,75 @@ function update_ensemble!(
...
+ - cov_threshold :: Threshold below which the reduction in covariance determinant results in a warning.
+ - Δt_new :: Time step to be used in the current update.
+ - deterministic_forward_map :: Whether output `g` comes from a deterministic model.
+ - failed_ens :: Indices of failed particles. If nothing, failures are computed as columns of `g`
+    with NaN entries.
+"""
+function update_ensemble!(
+    ekp::EnsembleKalmanProcess{FT, IT, Inversion},
+    g::AbstractMatrix{FT},
+    stoch_flag::Bool;
```
odunbar, line 262: I think this would be better as a `kwarg`. As it seems like this is something we will always be using in future, we may want the default to be `true` even?

ilopezgp: I agree, in general we should not pollute the required input fields with options like this.

costachris:
Does it make sense to perform inversions with both localization and parameter noise simultaneously? They both have the effect of breaking the subspace property of the initial ensemble.  I found these cases to be unstable in the tests.

src/EnsembleKalmanInversion.jl
costachris: good point - resolved in #236

docs/src/inflation.md
```diff
@@ -0,0 +1,45 @@
+# Inflation 
+Inflation is a method which systematically adds stochastic perturbations to the parameter ensemble in the form of gaussian noise. 
+Following the standard Kalman update, the parameter ensemble is perturbed in the following manner:
+```math
+    u_{n+1} = u_n + \zeta_{n} \qquad (1)
```
odunbar, line 5: This is just the additive inflation. Perhaps just write down that the mulitplicative inflation based on expanding the ensemble away from it's mean in a deterministic way, while additive inflation is based on adding this noise term?

docs/src/inflation.md
```diff
@@ -0,0 +1,45 @@
...
+```
+
+where `` \zeta_{n} `` is noise drawn from a gaussian distribution. Two distinct forms of inflation are implemented.
+For both implementations, a scaling factor ``s`` is included to extend usage to cases with mini-batching. For mini-batching,
```
odunbar, line 9: could you first mention that ``Delta t``, then mention the `s` in regards to this?

docs/src/inflation.md
```diff
@@ -0,0 +1,45 @@
...
+the scaling factor should be:
+```math
+    s = \frac{|B|}{|C|}
+```
+where `` |B| `` is the mini-batch size and `` |C| `` is the full dataset size, to account for sampling error.
```
odunbar, line 14: ", to account for sampling error" should be moved earlier

docs/src/inflation.md
```diff
@@ -0,0 +1,45 @@
...
+
+where `` \zeta_{n} `` is noise drawn from a gaussian distribution. Two distinct forms of inflation are implemented.
+For both implementations, a scaling factor ``s`` is included to extend usage to cases with mini-batching. For mini-batching,
+the scaling factor should be:
+```math
+    s = \frac{|B|}{|C|}
+```
+where `` |B| `` is the mini-batch size and `` |C| `` is the full dataset size, to account for sampling error. 
+
+## Multiplicative Inflation
```
odunbar, line 16: mention that multiplicative inflation does not change the span of the ensemble, but additive inflation does (in their respective sections)

docs/src/inflation.md
```diff
@@ -0,0 +1,45 @@
...
+Multiplicative effectively scales parameter vectors in parameter space, such that the perturbed
+ensemble remains in the linear span of the original ensemble. The implementated update equation follows
+[Huang et al, 2022](https://arxiv.org/abs/2204.04386) eqn. 41:
+
+```math
+\begin{aligned}
+    m_{n+1} = m_{n} ; \qquad u^{j}_{n + 1} = m_{n+1} + \sqrt{\frac{1}{1 - s \Delta{t}}} \left(u^{j}_{n} - m_{n} \right) \qquad (2)
+\end{aligned}
+```
+where ``m`` is the ensemble average which remains unchanged after the perturbation.
```
odunbar, line 26: In this way the variance of the ensemble is increased by ... while the mean remains fixed

docs/src/inflation.md
```diff
@@ -0,0 +1,45 @@
...
+```
+
+## Additive Inflation 
+Additive inflation breaks the linear subspace property, meaning the parameter ensemble can evolve
+outside of the span of the initial ensemble. In additive inflation, the update equation draws
+noise from the following distribution.
+
+```math
+    \zeta_{n} \sim N(0, \frac{s \Delta{t} }{1 - s \Delta{t}} C_n) \qquad (3)
+```
```
odunbar, line 40: this increases the variance by ... while the mean remains fixed

test/Inflation/runtests.jl
```diff
@@ -0,0 +1,112 @@
...
+
+# Read inverse problem definitions
+include("../EnsembleKalmanProcess/inverse_problem.jl")
+
+n_obs = 30                  # dimension of synthetic observation from G(u)
+ϕ_star = [-1.0, 2.0]        # True parameters in constrained space
+n_par = size(ϕ_star, 1)
+noise_level = 0.1           # Defining the observation noise level (std)
+N_ens = 5000                 # number of ensemble members
+N_iter = 5
```
odunbar, line 21: do we need 5 iteations here? I feel we just need 1 for the test, as we aren't comparing the solution.

test/Inflation/runtests.jl
```diff
@@ -0,0 +1,112 @@
...
+
+            # ensure inflation increases variance of ensemble
+            ϕ_final_standard = get_ϕ_final(prior, ekiobj)
+            ϕ_final_mult_inflation = get_ϕ_final(prior, eki_mult_inflation)
+            ϕ_final_add_inflation = get_ϕ_final(prior, eki_add_inflation)
+            ϕ_mean_final_standard = get_ϕ_mean_final(prior, ekiobj)
+            ϕ_mean_final_mult_inflation = get_ϕ_mean_final(prior, eki_mult_inflation)
+            ϕ_mean_final_add_inflation = get_ϕ_mean_final(prior, eki_add_inflation)
+            standard_update_variance = norm(ϕ_final_standard .- ϕ_mean_final_standard)
+            @test norm(ϕ_final_mult_inflation .- ϕ_mean_final_mult_inflation) > standard_update_variance
```
odunbar, line 96: could this be more precise? we have a large ensemble and we know what the variance inflation should be in each case?

odunbar (commented):
Hi @costachris this looks great - I have left comments mainly on tightening up the docs and tests. 

Sorry for some reason I added them as extra comments rather than a review!

PS could you also reference issue #230 in the PR

docs/src/inflation.md
```diff
@@ -0,0 +1,42 @@
+# Inflation 
+Inflation is a approach that slows ensemble collapse in Kalman methods.
+Two distinct forms of inflation are implemented, in which perturbations are applied immediately following the standard Kalman update. 
+Multiplicative inflation expands ensemble members away from their mean in a
+deterministic manner, whereas additive inflation hinges on the addition of stochastic noise to ensemble members.
+For both implementations, a scaling factor ``s`` is included to extend functionality to cases with mini-batching.
```
odunbar, line 6:
```suggestion

For both implementations, a scaling factor ``s`` is included to extend functionality to cases with mini-batching. 
```

docs/src/inflation.md
```diff
@@ -0,0 +1,42 @@
...
+```
+
+## Additive Inflation 
+Additive inflation is implemented by systematically adding stochastic perturbations to the parameter ensemble in the form of gaussian noise. Additive inflation breaks the linear subspace property, meaning the parameter ensemble can evolve outside of the span of the initial ensemble. In additive inflation, the ensemble is perturbed in the following manner after the standard Kalman update:
+
+```math
+     u_{n+1} = u_n + \zeta_{n} \qquad (2) \\
+    \zeta_{n} \sim N(0, \frac{s \Delta{t} }{1 - s \Delta{t}} C_n) \qquad (3)
+```
+This increases the variance across vector components by `` \frac{s \Delta{t} }{1 - s \Delta{t}} ``, while the mean remains fixed.
```
odunbar, line 38: Be careful when saying "increase", as above we **multiply** the variance by `1/(1-sDelt)` whereas here we **add** `sDelt/(1-sDelt)` to the variance.
Actually these give a new ensemble with the same variance right? Adding noise to each particle gives new variance scaled by `1 + sDelt/(1-sDelt) = (1-sDelt + sDelt) / (1-sDelt) = 1/(1-sDelt)` Might be worth clearing this up

odunbar (approved):
Looks great, 

I have 2 more tiny comments, and could you add the quick `get_localizer` in the `test/Localizers/runtests.jl` that codecov is complaining about. Then happy for merge

docs/src/inflation.md
odunbar: Oh we need C_n here

costachris:
@odunbar I included the option for using the prior covariance and updated the tests as discussed (including one for prior cov based noise). The docs now state more clearly how the covariance changes.

docs/src/inflation.md
```diff
@@ -0,0 +1,60 @@
+# Inflation 
+Inflation is a approach that slows ensemble collapse in Kalman methods.
```
ilopezgp, line 2:
```suggestion
Inflation is an approach that slows down collapse in ensemble Kalman methods.
```

docs/src/inflation.md
```diff
@@ -0,0 +1,60 @@
...
+Two distinct forms of inflation are implemented, in which perturbations are applied immediately following the standard Kalman update.
```
ilopezgp, line 3:
```suggestion
Two distinct forms of inflation are implemented in this package. Both involve perturbing the ensemble members following the standard update rule of the chosen Kalman process. 
```

docs/src/inflation.md
```diff
@@ -0,0 +1,60 @@
...
+ensemble remains in the linear span of the original ensemble. The implemented update equation follows
+[Huang et al, 2022](https://arxiv.org/abs/2204.04386) eqn. 41:
+
+```math
+\begin{aligned}
+    m_{n+1} = m_{n} ; \qquad u^{j}_{n + 1} = m_{n+1} + \sqrt{\frac{1}{1 - s \Delta{t}}} \left(u^{j}_{n} - m_{n} \right) \qquad (1)
+\end{aligned}
+```
+where ``m`` is the ensemble average. In this way,
+the parameter covariance is increased by a multiplicative factor of ``\frac{1}{1 - s \Delta{t}}``, while the mean remains fixed.
```
ilopezgp, line 25:
```suggestion
the parameter covariance is inflated by a factor ``\frac{1}{1 - s \Delta{t}}``, while the ensemble mean remains fixed.
```

docs/src/inflation.md
```diff
@@ -0,0 +1,60 @@
...
+     C_{n + 1} = \frac{1}{1 - s \Delta{t}} C_{n} \qquad (2)
+```
+
+Multiplicative inflation can be used by flagging the `update_ensemble!` method as follows:
+```julia
+    EKP.update_ensemble!(ekiobj, g_ens; multiplicative_inflation = true, s = 1.0)
+```
+
+## Additive Inflation 
+Additive inflation is implemented by systematically adding stochastic perturbations to the parameter ensemble in the form of gaussian noise. Additive inflation breaks the linear subspace property, meaning the parameter ensemble can evolve outside of the span of the initial ensemble. In additive inflation, the ensemble is perturbed in the following manner after the standard Kalman update:
```
ilopezgp, line 36:
```suggestion
Additive inflation is implemented by systematically adding stochastic perturbations to the parameter ensemble in the form of Gaussian noise. Additive inflation breaks the linear subspace property, meaning the parameter ensemble can evolve outside of the span of the initial ensemble. In additive inflation, the ensemble is perturbed in the following manner after the standard Kalman update:
```

docs/src/inflation.md
```diff
@@ -0,0 +1,60 @@
...
+
+```math
+     u_{n+1} = u_n + \zeta_{n} \qquad (3) \\
+    \zeta_{n} \sim N(0, \frac{s \Delta{t} }{1 - s \Delta{t}} C_n) \qquad (4)
+```
+This increases the parameter covariance by a multiplicative factor of ``\frac{1}{1 - s \Delta{t}}`` as in eqn. 2 , while the mean remains fixed.
```
ilopezgp, line 42:
```suggestion
This inflates the parameter covariance by a factor of ``\frac{1}{1 - s \Delta{t}}`` as in eqn. 2 , while the ensemble mean remains fixed.
```

src/EnsembleKalmanProcess.jl
```diff
@@ -448,20 +457,109 @@ function get_cov_blocks(cov::AbstractMatrix{FT}, p::IT) where {FT <: Real, IT <:
...
+
+    if scaled_Δt >= 1.0
+        error(string("Scaled time step: ", scaled_Δt, " is >= 1.0", "\nChange s or EK time step."))
+    end
+
+    if isnothing(prior)
+        Σ = get_u_cov_final(ekp)
+    else
+        Σ = cov(prior)
+    end
```
ilopezgp, line 518: Σ = isnothing(prior) ? get_u_cov_final(ekp) : cov(prior)

odunbar: To add to this, the prior is already stored in EKP, so 

So we could just pass in a flag, for example 
`use_prior_cov::Bool = false` 
and then here:
`Σ = use_prior_covariance ?  cov(get_u_prior(ekp)) : get_u_cov_final(ekp)`

I don't think we envisage other inflation distributions here, so I presume we do not need the ability to pass a generic one in

costachris: As far as I can tell `get_u_prior(ekp)` will get the initial parameter ensemble that is randomly drawn from the prior. So calling `cov(get_u_prior(ekp))` will give an approximation of the prior covariance; whereas calling `cov(prior)` (passing the prior distribution) will give the true prior covariance.  No sure if it makes a practical difference, but that's the only reason I went with passing the prior ParameterDistribution, but happy to switch to `cov(get_u_prior(ekp))` if that makes more sense.

odunbar: This is a good point, but using the ensemble to estimate the prior cov is still consistent with the default option (to use the current ensemble to estimate the current covariance) so I doubt there will be loss of accuracy here.

ilopezgp (approved):
LGTM

PR #236, opened 2022-11-20
Use multiple dispatch on update_ensemble calls

costachris:
## Purpose 
Use multiple dispatch on process-specific `update_ensemble!` calls, allowing a master `update_ensemble!` call to pass general keyword arguments.

### DIFF
--- a/src/EnsembleKalmanInversion.jl
+++ b/src/EnsembleKalmanInversion.jl
@@ -118,7 +118,8 @@ end
 """
     update_ensemble!(
         ekp::EnsembleKalmanProcess{FT, IT, Inversion},
-        g::AbstractMatrix{FT};
+        g::AbstractMatrix{FT},
+        process::Inversion;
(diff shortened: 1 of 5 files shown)


src/EnsembleKalmanInversion.jl
```diff
@@ -118,7 +118,8 @@ end
 """
     update_ensemble!(
         ekp::EnsembleKalmanProcess{FT, IT, Inversion},
-        g::AbstractMatrix{FT};
+        g::AbstractMatrix{FT},
+        process_type::Inversion;
```
odunbar, line 122: Note this is not the `process_type` it is the `process`

You have chosen to dispatch using an _object of type ::Inversion_ rather than just the _type ::Inversion_

Nothing wrong with this but we should change the variable name to `process`

odunbar (approved):
Could you correct the naming to `process`, in each function, then LGTM!

PR #249, opened 2023-01-26
Add dict-based transforms

### LINKED ISSUE #248: Transforming samples requires application to dicts
### Issue
We should add the functionality to transform parameter dicts, not just arrays. This should be straightforward as we must use the prior to perform the transform, and the prior also contains the parameter names.

Example setting: 
Currently, to transform samples from a `Samples` distribution, we must convert it into an array first with the correct ordering
```julia 
# S, Samples distribution (e.g. a posterior), 
# P, distribution (e.g. a prior with same parameter names)
samples_dict = get_distribution(S)
param_names = get_name(S)
samples_arr = zeros(sum([size(samples_dict[n],1) for n in param_names]), size(samples_dict[param_names[1]],2))
for (i,k) in enumerate(param_names)
        samples_arr[i,:] = samples_dict[k]
end
constrained_samples = transform_unconstrained_to_constrained(priors, samples_arr)
```

### Suggested solution

This is what we would want to transform samples from a `Samples` distribution

```julia
# S, Samples distribution (e.g. a posterior), 
# P, distribution (e.g. a prior with same parameter names)
samples_dict = get_distribution(S) #a Dict(param_name => samples_for_name)
constrained_samples_dict = transform_unconstrained_to_constrained(samples_dict)
```

Which can be done by adding the code snippet into functions `transform_constrained_to_unconstrained(pd::ParameterDistribution, d::Dict)`, `transform_unconstrained_to_constrained(pd::ParameterDistribution, d::Dict)` into `ParameterDistributions.jl`

odunbar:
## Purpose 

Closes #248 , by providing convenience methods to deal with the results of `get_distribution(Samples(...))`  

## Content

- add new `transform_unconstrained_to_constrained` for dicts
- add new `transform_constrained_to_unconstrained` for dicts
- add `get_all_constraints` with `return_dict` flag
- add `Samples`-based `ParameterDistribution` constructor from `AbstractMatrix` input
- add unit tests

----

### DIFF
--- a/src/ParameterDistributions.jl
+++ b/src/ParameterDistributions.jl
@@ -400,6 +400,23 @@ function ParameterDistribution(
 
 end
 
+"""
+    ParameterDistribution(distribution_samples::AbstractMatrix,
+                          constraint::Union{ConstraintType,AbstractVector{ConstraintType}},
+                          name::AbstractString;
+        params_are_columns::Bool = true)
+
+constructor of a Samples ParameterDistribution from a matrix `distribution_samples` of parameters stored as columns by defaut, (array of) `constraint`, `name`.
+"""
+function ParameterDistribution(
+    distribution_samples::AbstractMatrix,
+    constraint::Union{ConstraintType, AbstractVector},
+    name::AbstractString;
+    params_are_columns::Bool = true,
+)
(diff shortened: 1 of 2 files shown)


odunbar:
merging, as this simply adds functionality and tests, and is compatible with current interface.

PR #252, opened 2023-02-15
Add GaussianRandomFields to ParameterDistributions

### LINKED ISSUE #251: Gaussian Random Field parameters
This will enable parameters to be based in a function space, As a first stage we could look at representations using Gaussian Random Fields, using the package [`GaussianRandomFields.jl`](https://github.com/PieterjanRobbe/GaussianRandomFields.jl)

odunbar:
## Purpose 
- Addresses #251  
This PR will enable users to learn parameters that are discretizations of functions. 

## Content

- Currently allows: ND -> 1D functions with zero-mean allowed, stored flat. 
- hierarchy: `GaussianRandomField <: FunctionParameterDistributionType <: ParameterDistributionType`
- We use the package `GaussianRandomFields.jl` as the back-end
- We provide a default for the prior coefficients distribution under-the-hood, when given a GRF. Otherwise we allow a user-defined coefficient parameter distribution
- When built as a `ParameterDistribution`, the user provides a single constraint and this constraint is applied to all discretization points when the function is constructed and constrained.
- Compatible with the transform mappings: for parameter distributions. E.g the default is now, given coefficients vector `x` in unconstrained space. `u->c ` will transform the unconstrained to constrained coefficients, then builds the unconstrained function, then constrains this function
- Docstrings
- Darcy Flow Example in `example/Darcy` Note: We generate truth and solve problem with efunction expansion.

### Plots from Tests
1D Samples of the default prior distribution based on `GRF(Matern(smooth. 1.0,corr. len. 0.1))`
GRF_samples

A 2D sample with `bounded(-5,-3)` constraint
GRF_samples_constrained
The unconstrained 2D sample
GRF_samples_unconstrained

### Results from Darcy Flow:
**Prior Mean**
output_it_0
**Final Ensemble Mean**
output_it_10
**Truth**
output_true

----


Zhengyu-Huang (approved):
The results are amazing! Thanks for the great work!

PR #264, opened 2023-03-21
Quickly plot marginals for parameter distribution

### LINKED ISSUE #254: Plotting of parameter distribution (marginals) from a simple call
[StatsPlots](https://github.com/JuliaPlots/StatsPlots.jl/blob/master/README.md#distributions) provides support for julia distributions. 

It would be great to have a quick plotting functionality `plot(pd::ParameterDistribution)`

e.g. if pd was a combined [1D-dist, 3D multivariate dist, 2D histogram] it would plot all of the following
- a 1D pdf plot 
- a corner plot of the multivariate distribution, or three marginal plots
- a marginal histogram in each dimension, (or if 2D, a heatmap) 

Users can specify whether they want the plots in transformed or untransformed spaces.

odunbar:
## Purpose 

Closes #254

## Content

- For the wide variety of possible distributions, the easiest general solution is to plot marginal histograms of distributions, so this is the approach we take.
- Adds recipe for `plot(pd::ParameterDistribution)` 
- Adds recipe for `plot(pd <: ParameterDistributionType)` 
- Adds unit test which plots marginals (set `ENV["CES_TEST_PLOT_OUTPUT"] = true` prior to `] test`)
- Adds example into the priors docs page.

### In future
I expect these can be improved upon in various ways, using other plot-types, or only plotting unique distributions where they are repeated etc. but for this PR I just wanted something simple to work on any case.

## Example from `test/`

A complicated distribution: 
```julia
d1 = Parameterized(MvNormal(zeros(4), 0.1 * I))
c1 = [no_constraint(), bounded_below(-1.0), bounded_above(0.4), bounded(-0.1, 0.2)]
name1 = "constrained_mvnormal"
u1 = ParameterDistribution(d1, c1, name1)

d2 = Samples([1 2 3 4])
c2 = [bounded(10, 15)]
name2 = "constrained_sampled"
u2 = ParameterDistribution(d2, c2, name2)
            
d3 = VectorOfParameterized(repeat([Beta(2, 2)], 3))
c3 = repeat([no_constraint()], 3)
name3 = "vector_beta"
u3 = ParameterDistribution(d3, c3, name3)
            
u = combine_distributions([u1, u2, u3])
```

```julia
plot(u)
```

```julia
plot(u, constrained=false)
```

----

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -12,6 +12,7 @@ MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
 Optim = "429524aa-4258-5aef-a3af-852621145aeb"
 QuadGK = "1fd47b50-473d-5c70-9696-f719f8f3bcdc"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+RecipesBase = "3cdcf5f2-1ef4-517c-9805-6587b60abb01"
 SCS = "c946c3f1-0d1f-5ce8-9dea-7daa1f7e2d13"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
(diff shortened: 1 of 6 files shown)


odunbar:
NB: codecov doesn't like this because it only tests the PlotRecipes if plots are enabled (and github actions doesn't enable them)

odunbar:
Note that bors has succeeded on a rerun of the test suite, 

it seems like an unrelated test of inflation has a slightly too-tight tolerance which is subject to failure on random-seed change - see here for the refresh:
https://github.com/CliMA/EnsembleKalmanProcesses.jl/actions/runs/4493567718

sandreza (approved):
The changes look sensible to me! The only thing left is to let Bors do its thing

PR #280, opened 2023-05-02
removing find_ekp_stepsize

### LINKED ISSUE #271: Remove use of covariance determinant
The covariance determinant is used several times in the package; namely, there is a warning issued if it decreases too much, and it is also used in `find_ekp_stepsize` to find a stepsize. The first use will be removed by #250, but the second one should also be removed in my view, since the determinant is not a sensible quantity when the covariance is rank deficient (number of parameters > number of ensemble members). Perhaps the trace could be used instead?

eviatarbach:
Closes #271

### DIFF
--- a/docs/src/API/Inversion.md
+++ b/docs/src/API/Inversion.md
@@ -7,5 +7,4 @@ CurrentModule = EnsembleKalmanProcesses
 ```@docs
 Inversion
 eki_update
-find_ekp_stepsize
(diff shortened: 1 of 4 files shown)


odunbar (approved):
LGTM thanks! I also searched clima for `find_ekp_stepsize` and turned up nothing so there's a good chance noone is using it.

PR #281, opened 2023-05-04
pass RNG instead of seeds

### LINKED ISSUE #260: Should be passing an RNG here, not just a seed
https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/3e7494ee0f32db1ba2f7160e8c37bdceed34b2d1/src/EnsembleKalmanProcess.jl#L354

Just noticed that we pass in a seed here, this should be passing around an `rng::AbstractRNG` so users can use more stable options than `Random.seed!()`. 

In particular we should be writing functions that take in 
```function f(..., rng::RNG) where {RNG <: AbstractRNG}``` 
Or have a keyword with default 
```function f(..., rng = Random.default_rng())``` 

and the `rng` should be stored in the ekp object.

This will make randomness far more traceable in our code.

costachris:
Closes #260 

Pass RNG  objects instead of seeds to make randomness more traceable.

### DIFF
--- a/examples/AerosolActivation/aerosol_activation.jl
+++ b/examples/AerosolActivation/aerosol_activation.jl
@@ -39,7 +39,9 @@ using Plots
 using Distributions
 using LinearAlgebra
 using Random
+
 rng_seed = 44
+rng = Random.seed!(Random.GLOBAL_RNG, rng_seed)
 
 using EnsembleKalmanProcesses
 using EnsembleKalmanProcesses.ParameterDistributions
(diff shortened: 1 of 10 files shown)


odunbar:
@costachris  Could you rebase with main, before I review, hopefully then all tests will pass! Cheers

costachris:
@odunbar The EKP test is failing because we're doing two sequential calls to `construct_initial_ensemble`, passing the same rng object without reseeding, and checking that they're equal. This test made sense when we were passing the seed to `construct_initial_ensemble`, but that's no longer the case. The simple solution is to reseed the rng object right before the second call to `construct_initial_ensemble` (outside of the function), but just want to double-check that's the behavior we want.

odunbar (approved):
Looks good, Thank you! Regarding your comment. I think we should just remove that test. 

As you say, the reproducibility should not occur unless the user does something external now - by explicitly copying/reseeding the `rng`, or the `GLOBAL_RNG` outside the function. This is then something already tested in Random/Distributions etc. so not our concern any more.

Once this is done, please merge!

PR #283, opened 2023-05-05
Fast concatentation

### LINKED ISSUE #274: remove cat(X...) statements
For convenience we often use `cat(X..., dims=p)` statements to concatenate vectors and matrices. During some profile I see these become terribly inefficient at even moderate problems. We should probably remove them. As an example"

```julia
X = [zeros(3,3) for i = 1:1000]

# option 1 
xx = cat(X...,dims=3)

# option 2
function squidge(X::AbstractVector) 
       xx = zeros(size(X[1],1),size(X[1],2),length(X))
       for i = 1:length(X)
           xx[:,:,i] = X[i]
       end
       return xx
end
xx = squidge(X)
```
option 1 gives e.g.: 
 `0.037136 seconds (10.01 k allocations: 13.965 MiB)`
option 2 gives e.g.:
`0.000088 seconds (2 allocations: 70.375 KiB)`

## Action
Replace `cat(X...,dims=p)` with the appropriate loops.

haakon-e:
closes #274

### DIFF
--- a/src/ParameterDistributions.jl
+++ b/src/ParameterDistributions.jl
@@ -369,7 +369,8 @@ function ParameterDistribution(param_dist_dict::Union{Dict, AbstractVector})
 
     # flatten the structure
     distribution = getindex.(param_dist_dict_array, "distribution")
-    flat_constraint = cat(getindex.(param_dist_dict_array, "constraint")..., dims = 1)
+    flat_constraint = reduce(vcat, getindex.(param_dist_dict_array, "constraint"))
+    flat_constraint = isa(flat_constraint, Vector) ? flat_constraint : [flat_constraint]
     name = getindex.(param_dist_dict_array, "name")
 
     # build the object
@@ -412,11 +413,10 @@ function ParameterDistribution(
 
     # flatten the structure
     distribution_vec = [distribution]
-    flat_constraint_vec = cat(constraint_vec..., dims = 1)
     name_vec = [name]
 
     # build the object
-    return ParameterDistribution(distribution_vec, flat_constraint_vec, name_vec)
+    return ParameterDistribution(distribution_vec, constraint_vec, name_vec)
 
 end
 
@@ -460,9 +460,9 @@ Form a ParameterDistribution by concatenating a vector of single ParameterDistri
 """
 function combine_distributions(pd_vec::AbstractVector{PD}) where {PD <: ParameterDistribution}
     # flatten the structure
-    distribution = cat([d.distribution for d in pd_vec]..., dims = 1)
-    constraint = cat([d.constraint for d in pd_vec]..., dims = 1)
-    name = cat([d.name for d in pd_vec]..., dims = 1)
+    distribution = reduce(vcat, getfield.(pd_vec, :distribution))
+    constraint = reduce(vcat, getfield.(pd_vec, :constraint))
+    name = reduce(vcat, getfield.(pd_vec, :name))
     return ParameterDistribution(distribution, constraint, name)
 end
 
@@ -572,7 +572,7 @@ parameters as columns. `rng` is optional and defaults to `Random.GLOBAL_RNG`. `n
 optional and defaults to 1. Performed in computational space.
 """
 function sample(rng::AbstractRNG, pd::ParameterDistribution, n_draws::IT) where {IT <: Integer}
-    return cat([sample(rng, d, n_draws) for d in pd.distribution]..., dims = 1)
+    return reduce(vcat, sample.(rng, pd.distribution, n_draws))
 end
 
 # define methods that dispatch to the above with Random.GLOBAL_RNG as a default value for rng
@@ -669,17 +669,15 @@ end
 
 function logpdf(pd::ParameterDistribution, xarray::AbstractVector{FT}) where {FT <: Real}
     #first check we don't have sampled distribution
-    for d in pd.distribution
-        if typeof(d) <: Samples
-            throw(
-                ErrorException(
-                    "Cannot compute logpdf of Samples distributions. Consider using a Parameterized type for your prior.",
-                ),
-            )
-        end
+    if any(isa.(pd.distribution, Samples))
+        throw(
+            ErrorException(
+                "Cannot compute logpdf of Samples distributions. Consider using a Parameterized type for your prior.",
+            ),
+        )
     end
     #assert xarray correct dim/length
-    if size(xarray)[1] != ndims(pd)
+    if length(xarray) != ndims(pd)
         throw(
             DimensionMismatch(
                 "xarray must have dimension equal to the parameter space. Expected $(ndims(pd)), got $(size(xarray)[1])",
@@ -691,7 +689,7 @@ function logpdf(pd::ParameterDistribution, xarray::AbstractVector{FT}) where {FT
     batches = batch(pd)
 
     # perform the logpdf of each of the distributions, and returns their sum    
-    return sum(cat([logpdf(d, xarray[batches[i]]) for (i, d) in enumerate(pd.distribution)]..., dims = 1))
+    return sum(sum(logpdf(d, xarray[batches[i]])) for (i, d) in enumerate(pd.distribution))
 end
 
 #extending StatsBase cov,var
@@ -702,23 +700,13 @@ Returns a flattened variance of the distributions
 var(d::Parameterized) = var(d.distribution)
 var(d::Samples) = var(d.distribution_samples, dims = 2)
 function var(d::VectorOfParameterized)
-    d_dims = get_dimensions(d)
-    block_var = Array{Any}(undef, size(d_dims)[1])
-
-    for (i, dimension) in enumerate(d_dims)
-        block_var[i] = var(d.distribution[i])
-    end
-    return cat(block_var..., dims = 1)
+    block_var = var.(d.distribution)
+    return reduce(vcat, block_var)
 end
 
 function var(pd::ParameterDistribution)
-    d_dims = get_dimensions(pd)
-    block_var = Array{Any}(undef, size(d_dims)[1])
-
-    for (i, dimension) in enumerate(d_dims)
-        block_var[i] = var(pd.distribution[i])
-    end
-    return cat(block_var..., dims = 1) #build the flattened vector
+    block_var = var.(pd.distribution)
+    return reduce(vcat, block_var) #build the flattened vector
 end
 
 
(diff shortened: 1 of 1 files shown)


haakon-e:
Looking at the actual `cat`-statements we have in the code, I realized good solutions actually are different depending on details about the vectors/matrices are being reduced. So I will test a bit for the different types to verify which patterns repeat themselves.

src/ParameterDistributions.jl
```diff
@@ -369,7 +369,8 @@ function ParameterDistribution(param_dist_dict::Union{Dict, AbstractVector})
 
     # flatten the structure
     distribution = getindex.(param_dist_dict_array, "distribution")
-    flat_constraint = cat(getindex.(param_dist_dict_array, "constraint")..., dims = 1)
+    flat_constraint = reduce(vcat, getindex.(param_dist_dict_array, "constraint"))
+    flat_constraint = isa(flat_constraint, Vector) ? flat_constraint : [flat_constraint]
```
haakon-e, line 373: Replace `cat(cons..., dims=1)` by `reduce(vcat, cons)`. This is generally faster and less allocating. In the case where `cons` is a 1-vector, the `reduce` operation gives a `Constraint` instead of a `Vector{Constraint}`, therefore L373 is needed to ensure we end up with a vector.

src/ParameterDistributions.jl
```diff
@@ -412,11 +413,10 @@ function ParameterDistribution(
 
     # flatten the structure
     distribution_vec = [distribution]
-    flat_constraint_vec = cat(constraint_vec..., dims = 1)
     name_vec = [name]
 
     # build the object
-    return ParameterDistribution(distribution_vec, flat_constraint_vec, name_vec)
+    return ParameterDistribution(distribution_vec, constraint_vec, name_vec)
```
haakon-e, line 419: I think the input arguments to this function ensures that `constraint_vec` is already a flat vector, so `cat(..., dims=1)` shouldn't be needed.

src/ParameterDistributions.jl
```diff
@@ -669,17 +669,15 @@ end
...
-                    "Cannot compute logpdf of Samples distributions. Consider using a Parameterized type for your prior.",
-                ),
-            )
-        end
+    if any(isa.(pd.distribution, Samples))
+        throw(
+            ErrorException(
+                "Cannot compute logpdf of Samples distributions. Consider using a Parameterized type for your prior.",
+            ),
+        )
```
haakon-e, line 677: replace loop by broadcasting and any. Does not change performance, but personally is a bit more readable.

src/ParameterDistributions.jl
```diff
@@ -669,17 +669,15 @@ end
...
     end
     #assert xarray correct dim/length
-    if size(xarray)[1] != ndims(pd)
+    if length(xarray) != ndims(pd)
```
haakon-e, line 680: `xarray` is an `AbstractVector`, so `size(xarray)[1]` is equivalent to its length.

src/ParameterDistributions.jl
```diff
@@ -691,7 +689,7 @@ function logpdf(pd::ParameterDistribution, xarray::AbstractVector{FT}) where {FT
     batches = batch(pd)
 
     # perform the logpdf of each of the distributions, and returns their sum    
-    return sum(cat([logpdf(d, xarray[batches[i]]) for (i, d) in enumerate(pd.distribution)]..., dims = 1))
+    return sum(sum(logpdf(d, xarray[batches[i]])) for (i, d) in enumerate(pd.distribution))
```
haakon-e, line 692: Here I sum over each `logpdf` evaluation for each batch, then I sum the summed quantities.
The inner sum replaces the need for `cat(..., dims=1)`. Faster and less allocating

src/ParameterDistributions.jl
```diff
@@ -702,23 +700,13 @@ Returns a flattened variance of the distributions
...
 function var(d::VectorOfParameterized)
-    d_dims = get_dimensions(d)
-    block_var = Array{Any}(undef, size(d_dims)[1])
-
-    for (i, dimension) in enumerate(d_dims)
-        block_var[i] = var(d.distribution[i])
-    end
-    return cat(block_var..., dims = 1)
+    block_var = var.(d.distribution)
+    return reduce(vcat, block_var)
```
haakon-e, line 704: broadcast `var` instead of looping. Then `reduce(vcat, ...)` instead of `cat(..., dims=1)`.

src/ParameterDistributions.jl
```diff
@@ -702,23 +700,13 @@ Returns a flattened variance of the distributions
...
 function var(pd::ParameterDistribution)
-    d_dims = get_dimensions(pd)
-    block_var = Array{Any}(undef, size(d_dims)[1])
-
-    for (i, dimension) in enumerate(d_dims)
-        block_var[i] = var(pd.distribution[i])
-    end
-    return cat(block_var..., dims = 1) #build the flattened vector
+    block_var = var.(pd.distribution)
+    return reduce(vcat, block_var) #build the flattened vector
```
haakon-e, line 709: broadcast `var` instead of looping. Then `reduce(vcat, ...)` instead of `cat(..., dims=1)`.

src/ParameterDistributions.jl
```diff
@@ -771,71 +759,51 @@ Returns a concatenated mean of the parameter distributions.
...
 mean(d::Parameterized) = mean(d.distribution)
 mean(d::Samples) = mean(d.distribution_samples, dims = 2)
-function mean(d::VectorOfParameterized)
-    return cat([mean(dd) for dd in d.distribution]..., dims = 1)
-end
-function mean(pd::ParameterDistribution)
-    return cat([mean(d) for d in pd.distribution]..., dims = 1)
-end
+mean(d::VectorOfParameterized) = reduce(vcat, mean.(d.distribution))
+mean(pd::ParameterDistribution) = reduce(vcat, mean.(pd.distribution))
```
haakon-e, line 763: broadcast `mean` instead of looping. Then `reduce(vcat, ...)` instead of `cat(..., dims=1)`.

src/ParameterDistributions.jl
```diff
@@ -771,71 +759,49 @@ Returns a concatenated mean of the parameter distributions.
...
+    end
     for (name, idxs, d) in zip(param_names, pd_batch_idxs, pd.distribution)
-        push!(ret, Array(transform_constrained_to_unconstrained(d, pd_constraints[name], x[idxs, :])))
-    end #returns a list of row-blocks that are concatenated
-
-    return cat(ret..., dims = 1)
-
+        view(x_out, idxs, :) .= transform_constrained_to_unconstrained(d, pd_constraints[name], view(x, idxs, :))
+    end
+    return x_out
```
haakon-e, line 801: (this part is best viewed in "split" diff view, or just the new code).
In the old code, we created an empty array `ret = []` and pushed to this through a loop, then `cat(ret..., dims=1)` to  get a 1-d vector as output.

Here, I exploit the fact that we know the size of the output array to create an `x_out` of the correct size and type based on the input `x`. There is a slight complication if `pd` is a function distribution, in which case the output dimension is the sum of `n_eval_pts` for each individual distribution.
The second dimension `length(axes(x, 2))` equates to `1` if `x` is a Vector (i.e. 1 sample), or to the number of samples (i.e. number of columns) otherwise. 
That fact, combined with the fact that the new code which now writes _to_ the `x_out` through broadcasting (`.=`) and `view(x, idxs, :)` (which works for vectors and matrices) means that there is no longer a need for separate methods for `x::Vector` and `x::Matrix`.

src/ParameterDistributions.jl
```diff
@@ -873,87 +839,51 @@ function transform_constrained_to_unconstrained(
...
+    x::AbstractArray{FT};
     kwargs...,
 ) where {FT <: Real, PDT <: ParameterDistributionType}
-    return Array(hcat([c.unconstrained_to_constrained.(x[i, :]) for (i, c) in enumerate(constraints)]...)')
+    x_out = similar(x)
+    for (out, in, c) in zip(eachrow(x_out), eachrow(x), constraints)
+        out .= c.unconstrained_to_constrained.(in)
+    end
+    return x_out
 end
```
haakon-e, line 854: much of the explanation above for `transform_constrained_to_unconstrained` applies here to. However, since `ParameterDistributionType` does not include function distributions, the size of `x_out` is the same as `x`.

src/ParameterDistributions.jl
```diff
@@ -873,87 +839,51 @@ function transform_constrained_to_unconstrained(
...
-
-    return cat(ret..., dims = 1)
+    x_out = similar(x, x_out_size)
+    for (name, eval_idx, pd_idxs, d) in zip(param_names, eval_batch_idxs, pd_batch_idxs, pd.distribution)
+        view(x_out, eval_idx, :) .=
+            transform_unconstrained_to_constrained(d, pd_constraints[name], view(x, pd_idxs, :); build_flag)
+    end
+    return x_out
 
 end
```
haakon-e, line 888: Again, much of the discussion is the same as for `transform_constrained_to_unconstrained`, above.
There is one added complication however: The `dof` batch is different from the `eval` batch. In other words, the elements of `x` _may_ be interpreted as coefficients of a certain function as opposed to evaluation points. As such, we need to fetch the correct output size (`x_out_size`) and indices (`eval_batch_idxs`) that match the _evaluation_ size.

haakon-e (commented):
@odunbar I have added explanations to all my changes as review comments. Feel free to reply directly to each one you may have questions about.

For every change I made, I tested that the refactored function is faster and allocates less than the baseline. In many cases, the speedup is very large. In other cases, it's minimal.
Note: I've only tested this using structs I found in the unit/integration tests, which are typically very small. So it's conceivable, although unlikely, that the improvement for "real world" examples (i.e. very large distributions) may be different.

src/ParameterDistributions.jl
odunbar: again I would rather we call ndims with kwarg `evals` than call `n_eval_pts` directly

odunbar (commented):
Thanks a lot for this! It looks greatly improved. I do not think any of these changes will haunt us computationally compared with the use of `cat(...)`.

I've left a couple of small comments regarding using the `ndims` for function distributions. An once resolved I'm happy!

src/ParameterDistributions.jl
haakon-e: done.

src/ParameterDistributions.jl
haakon-e: done.

odunbar (approved):
Hi @haakon-e these changes look good! I see a test failed though. 

Could this be resolved - then I'm happy for merge.

haakon-e:
> Hi @haakon-e these changes look good! I see a test failed though.
> 
> Could this be resolved - then I'm happy for merge.

@odunbar It appears the test only fails on v1.6.
I tried to go about reproducing locally, but I am for some reason not able to even instantiate EKP.jl on v1.6.7 (I get error ``ERROR: expected package `Distributions [31c24e10]` to be registered``), so I'm at a loss for how start addressing this.

haakon-e:
It turns out that on `julia <= 1.8` the method `mapslices`, which we use to transform columns of a matrix, errors if the function that is applied to a column returns a `n x 1` matrix instead of a `n-vector`. For example:

```julia
# julia v1.8.4
julia> xx = zeros(4,5);

julia> mapslices(x -> reshape(x, 4, 1), xx; dims=1)
ERROR: DimensionMismatch: tried to assign 2 elements to 1 destinations
Stacktrace:
 [1] throw_setindex_mismatch(X::Vector{Base.OneTo{Int64}}, I::Tuple{Int64})
   @ Base ./indices.jl:191
 [2] setindex_shape_check
   @ ./indices.jl:245 [inlined]
 [3] setindex!(A::Vector{Base.OneTo{Int64}}, X::Vector{Base.OneTo{Int64}}, I::Vector{Int64})
   @ Base ./array.jl:973
 [4] mapslices(f::var"#5#6", A::Matrix{Float64}; dims::Int64)
   @ Base ./abstractarray.jl:2873
 [5] top-level scope
   @ REPL[5]:1
```

In contrast, this works in julia v1.9:
```julia
# julia v1.9.2
julia> xx = zeros(4,5);

julia> mapslices(x -> reshape(x, 4, 1), xx; dims=1)
4×5 Matrix{Float64}:
 0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0
```

I have updated my implementation of `transform_(un)constrained_to_(un)constrained` to return a vector if a vector is given, and a matrix if a matrix is given.
This does two things:
1. It fixes the aforementioned issue since `mapslices` passes the slices as vectors to the function
2. this improves type inference on both julia v1.8 and julia v1.9. In 1.8, julia failed to infer the return type of the method (`@code_warntype` indicated `Any`). In 1.9, julia was only able to say `Union{Matrix, Vector}` for matrix or vector input types.

All of this being said, as far as I can see the use of `mapslices` in the first place is not necessary. The following calls are equivalent ([see callsite in tests](https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/98ffabf200115a7207c294d49b16c72b46c7d396/test/ParameterDistributions/runtests.jl#L739))
```julia
x_real_constrained1 = mapslices(x -> transform_unconstrained_to_constrained(u1, x), x_unbd[1:4, :]; dims = 1)
x_real_constrained1 == transform_unconstrained_to_constrained(u1, x_unbd[1:4, :])
true
```
but this could be changed in a different PR

PR #285, opened 2023-05-09
Timesteppers for EKP, from ParameterEstimocean

### LINKED ISSUE #154: Add LearningRateSchedulers
Following typical optimizer structures in deep learning (tensor flow, pytorch), we can add the possibility of defining a `LearningRateScheduler` at construction time for EKP. 

Then, we would set the default dt based on the scheduler, and perhaps still expose this parameter for the user in case they want to try out some other schedule.

### LINKED ISSUE #253: Timestepping in EKP
Movement of timesteppers from [ParameterEstimocean](https://github.com/CliMA/ParameterEstimocean.jl/blob/main/src/PseudoSteppingSchemes.jl) into the EKP repository, in a way that is compatible with existing tools, and can include new timesteppers as they are developed.

odunbar:
Coauthored with @eviatarbach 

## Purpose 

Closes #253 
Closes #154 
## Content

- Implemented informative warning message for building 
`EnsembleKalmanProcess(...; Deltat = x, ...)` 
In favour of the new constructor
`EnsembleKalmanProcess(...; scheduler = DefaultScheduler(x), ...)`
and updated all unit tests to the latter.
- New `LearningRateScheduler` type and Objects in backwards compatible way, by adding the keyword `scheduler = ...`
    1. `DefaultScheduler(k)` (always does step `k` unless user overrides at a given step), 
    2. `MutableScheduler()` (initially does step `1`, until user overrides with `k`, then does `k` thereafter until user next overrides)
    3. `EKSStableScheduler()` - performs the stable EKS step calculation externally, in place of the original internal one.
    4. `DataMisfitController()` - based on Iglesias Yan 2021's Bayesian tempering scheme.    
- New Defaults are `DefaultScheduler(1)` usually, and `EKSStableScheduler` for EKS.
- Current examples no longer have `Deltat= x` expressions in constructors of EKP object.
- `calculate_timestep!` can output a flag if a termination condition is met, this will prevent further EKP updates (e.g. for `DataMisfitController`
- Further options for DMC when it reaches the termination condition. `on_terminate = stop` `continue` or `continue_fixed` to either terminate, continue, or continue with latest timestep fixed.
- Example comparing different schedulers (see results below)
- Docstrings & docs page, EKI and UKI example: [SEE DOCS HERE](https://clima.github.io/EnsembleKalmanProcesses.jl/previews/PR285/learning_rate_scheduler/)

## Also Changed
- A test coding pattern, where the forward map `G` transformed parameters internally, rather than using the parameter distributions to to return "constrained parameters" which is the proper way these should be used
- The Nonlinear forward map that had a multiplicative noise internal variability, now has a no-noise and additive noise version used in testing (a more appropriate problem for EKP).
 - UKI initialization (/bugfix?) for timestep = `0.0` not `\Delta t`. Allows for constructor cleanup and clean interface with new schedulers.
 - UKI now is consistent with other tools in building `Unscented{FT,IT}` types, not unparameterized `Unscented` types.

### Example experiment - explained more clearly in docs
Learning `(A,B)` from `f = exp(Asin(t) + B)`, from observing only `max(f)` and `mean(f)`

Final model solutions, as well as the max and mean of the solution 
left: black = true model (not the observed data)
right: black = observed truth, grey = mean/2*sd of true model observations
ensembles

### Individual experiment behaviour
mean ensemble member error to truth, (error), and mean ensemble member error to ensemble mean (spread).
Avoiding ensemble collapse => dashed and solid lines are similar. 
error_vs_spread_over_iteration

Green - constant step 0.02 for 50 steps (finishes at T=1)
Orange - constant step 0.5 for 50 steps (finishes at T=25)
Red - adaptive step, usually used for EKS (finishes at T=10^13?)
Purple - adaptive DMC step terminates at 12 steps (finishes at T=1)
Brown - adaptive DMC step with `on_terminate = continue` (finishes at T = 10^15)

### Average convergence over 100 experiments
50 ensemble members and 50 iterations (DMC typically terminates ~14-20 iterations)
Method      : DefaultScheduler{Float64}(0.5)
Final misfit: 0.011919044450605893
Final error : 0.027449304379020753
Final spread: 0.0031725385439674517
 
Method      : DefaultScheduler{Float64}(0.02)
Final misfit: 1.0919040792305175
Final error : 0.09912866440799409
Final spread: 0.07396181868428042
 
Method      : EKSStableScheduler{Float64}(1.0, 2.220446049250313e-16)
Final misfit: 5.648615687716918e-5
Final error : 0.03018464453288676
Final spread: 1.9601185362422715e-5
 
Method      : DataMisfitController{...}(..., "stop")
Final misfit: 1.1677896425071728
Final error : 0.09817906335167692
Final spread: 0.08739534124484036
 
Method      : DataMisfitController{...}(..., "continue")
Final misfit: 8.52533406253277e-17
Final error : 0.030283499635998162
Final spread: 5.974713527239014e-17

### Conclusion? 
These results imply that 
- DMC behaves similarly to fixed small-step EKI on `0->1`, but is far more efficient with regards to step size
- DMC and fixed small-step EKI on `0->1` both retain better spread in the ensemble than other options,
- Termination at T=1 (even averaged over 100's of runs) seems to give only conservative estimates of the mean, to an order of magnitude or more. (Likely to protect the spread of the ensemble?) 
- Continuing DMC causes it to mimic EKSStableScheduler behaviour but with less steps. (Brown is similar to red)

----


odunbar:
Note tests failing due to upgrades to Julia v1.9.0 - they do pass in v1.8.5. This will be addressed shortly

odunbar:
#286

docs/src/learning_rate_scheduler.md
```diff
@@ -0,0 +1,69 @@
...
+
+![Error vs spread](assets/error_vs_spread_over_iteration_compare_scheduler.png)
+
+*Left: the error and spread of the different timesteppers at over iterations of the algorithm for a single run.
+Right: the error and spread of the different timesteppers at their final iterations, averaged from 100 different draws of initial conditions.*
+
+From this plot there are several take-home messages.
+ - Termination at ``T=1``. DMC with termination (purple), closely mimics a small-fixed timestep (green) that finishes stepping at ``T=1``. Both retain more spread than other approaches, and DMC is far more efficient, typically terminating after around 10-20 steps, where fixed-stepping takes 50. We see that (for this experiment) this is a conservative estimate, as continuing to solve (e.g. brown) until later times often leads to a better error while still retaining similar "error vs spread" curves (before inevitable collapse).
+ - Large fixed step (orange). This is very efficient, but can get stuck in local minima when too large, it collapses the ensemble. Note that even on average it gives lower error to the true parameters than DMC.
```
eviatarbach, line 53: I think that getting stuck in local minima is usually a problem for small timesteps. For large timesteps the problem I think is usually "overshooting" the minimum.

docs/src/learning_rate_scheduler.md
```diff
@@ -0,0 +1,69 @@
...
+ - Termination at ``T=1``. DMC with termination (purple), closely mimics a small-fixed timestep (green) that finishes stepping at ``T=1``. Both retain more spread than other approaches, and DMC is far more efficient, typically terminating after around 10-20 steps, where fixed-stepping takes 50. We see that (for this experiment) this is a conservative estimate, as continuing to solve (e.g. brown) until later times often leads to a better error while still retaining similar "error vs spread" curves (before inevitable collapse).
+ - Large fixed step (orange). This is very efficient, but can get stuck in local minima when too large, it collapses the ensemble. Note that even on average it gives lower error to the true parameters than DMC. 
+ - Continuation ``T \to \infty``. Both EKSStable and DMC with continuation schedulers, perform very similarly. Both retain good ensemble spread for during convergence, and collapse after finding a local optimum. This optimum on average has the best error to the true parameters in this experiment. They appear to consistently find the same optimum as ``T\to\infty`` but DMC finds this in fewer iterations.
```
eviatarbach, line 54: "for during" -> "during"

docs/src/learning_rate_scheduler.md
```diff
@@ -0,0 +1,69 @@
...
+ - Adaptivity. The fixed-step may eventually find the same optimum as EKS/DMC, but after 50 iterations ``T\sim 10^{12} - 10^{15}`` in the adaptive algorithms, while the fixed stepper has reached ``T=25``. On the other end of the scale, the initial timesteps selected are often on the order of ``\Delta t \sim 10^{-8} - 10^{-6}``. The implication is that adaptivity is likely very useful for efficiency of these algorithms.
```
eviatarbach, line 55: I am not sure I understand how the efficiency follows from the discrepancy in times.

eviatarbach (changes requested):
Great work on this! I made a few comments in the code. I think the main issues are lack of tests for UKI and sparse EKI.

docs/src/learning_rate_scheduler.md
odunbar: done

docs/src/learning_rate_scheduler.md
odunbar: Hmm, I wonder if the intuition is backwards for EKI, or maybe our terminology is not clear enough? When the timestep is too large, I might obtain an overly-collapsed ensemble at some value, and it does not move far from this in future iterations. For small steps, I don't think I have observed any "getting stuck" in local minima, may this is due to the noise added in every iteration.

eviatarbach: Ah yes, that's a good point, I guess it's hard to apply the intuition from gradient descent. This would be interesting to think more about in the future, but we can leave it as you phrased it then.

src/LearningRateSchedulers.jl
```diff
@@ -0,0 +1,302 @@
...
+
+    q = maximum((M / (2 * Φ_mean), sqrt(M / (2 * Φ_var))))
+
+    if sum_Δt < 1
+        Δt = minimum([q, 1 - sum_Δt])
+    else # when termination condition satisfied but choose to continue
+        Δt = q
+    end
+
+    # in theory the following should be the same.
```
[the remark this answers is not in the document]
odunbar: removed.

docs/src/learning_rate_scheduler.md
odunbar: The adaptive methods reached their minimum value at `N_{it} = 20` or so, this  corresponded to a  very large algorithm time `T>>1`. It appears likely that the fixed timestepper will reach this same minimum but will take in `N_{it} >> 20` as it cannot take larger steps. So they are very efficient.

I can rewrite this to explain it better

docs/src/learning_rate_scheduler.md
```diff
@@ -0,0 +1,70 @@
...
+There are two termination times that the theory indicates are useful
+- ``T=1``: In the linear Gaussian case, the \{\theta_{N_\mathrm{it}}\} will represent the posterior distribution. In nonlinear cases we often find ``\bar{\theta}_{N_\mathrm{it}}`` (the ensemble mean) provides a conservative estimate of the true parameters, while retaining spread. In other words, it may be that one can obtain a more concentrated posterior around 
+- ``T\to \infty``: This ensures ensemble collapse. In this case we often observe a better estimate ``\bar{\theta}_{N_\mathrm{it}}`` to the true parameters, with no/little spread.
+
+## The experiment
+
+The experiment assesses the schedulers robustly by solving the inverse problem 100 times with different random seeds.
+Shown below is an example plot of one of these solves with each timestepper.
+
+![Solution ensembles](assets/ensemble_compare_scheduler.png)
```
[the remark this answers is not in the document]
odunbar: I've replaced it with the solution of one instance (i.e the model at the final ensemble mean) for one of the ICs. If this still isn't useful I'm happy to remove it!

test/EnsembleKalmanProcess/runtests.jl
```diff
@@ -61,6 +61,114 @@ inv_problems = [inv_problems..., nl_inv_problem]
...
+    @test dmclrs1.on_terminate == "stop"
+    dmclrs2 = EKP.DataMisfitController(on_terminate = "continue")
+    @test dmclrs2.on_terminate == "continue"
+    dmclrs3 = EKP.DataMisfitController(on_terminate = "continue_fixed")
+    @test dmclrs3.on_terminate == "continue_fixed"
+
+
+
+
+    # build EKP and eki objects
```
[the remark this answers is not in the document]
odunbar: I have added both sets of tests. The Sparse inversion tests are in the SparseInversion/ test directory.

Currently UKI tests with the `DataMisfitController` do not appear to converge, I will investigate later what is going on.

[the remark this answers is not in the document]
odunbar: I've added UKI example in the docs too now, and here it seems to perform ok. I wonder if there is some quirk in the test setup that's giving it problems

odunbar:
Leaving UKI alone, as trying to make easier tests for it still seems to cause poor performance, I suspect a bug rather than anything to do with schedulers.  Commenting out UKI test, to be uncommented when performance has been thoroughly assessed.

Created #287

eviatarbach:
Looks great to me! Thank you for all your work on this. Not sure why a test is failing on Windows but I think this should be ready to merge after that passes.

odunbar:
bors try

odunbar:
The windows test passed on rerun - I'll give it a couple more times to see if it is recurring

odunbar:
bors try

PR #286, opened 2023-05-17
Fix v1.9.0 testing

odunbar:
## Content

- Adds >=v1.9 tests where behaviour changes. 
Changes weres isolated to Failure handler tests, as the most up-to-date `LinearAlgebra` no longer allows as much propagation of NaNs, rather it throws errors. (I think this is a good idea)
- Removes the various precompile warnings about "FT defined but not used" 
- Removes the precompile warning of a clash in `sample` by now importing `GaussianRandomFields` rather that using it. This is (I think) because GRF.jl does not extend the StatsBase sample function, and we cant `import StatsBase: sample` AND `using GaussianRandomFields` without warnings.
- Update github workflows for docs and formatter to include `v1.X` from pinning to `1.7` 

----

### DIFF
--- a/.github/workflows/Docs.yml
+++ b/.github/workflows/Docs.yml
@@ -27,7 +27,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@v1
         with:
-          version: '1.7'
+          version: '1'
           show-versioninfo: 'true'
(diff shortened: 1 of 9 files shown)


odunbar:
Merge without review, when all checks pass for both < v1.9.0 and v1.9.0

PR #291, opened 2023-06-07
Allow users to select termination time for DataMisfitController

### LINKED ISSUE #290: We can improve DMC consistency by controlling the termination tolerance
Currently it is defaulted to terminate at T = 1, or we let it continue for K steps. To allow for more easy comparison/consistency with multiple experiements, it would be good to add the option setting a termination time of `T` for some `T>1`. E.g. One can set that an experiment suite all terminate at T=20, allowing for different numbers of adaptive steps in each experiment.

odunbar:
## Purpose 

Closes #290 

## Content

- Added a keyword argument `terminate_at` for DMC
- Added tests to show it stops at the correct time
- Added a line into the docs 

----

### DIFF
--- a/docs/src/learning_rate_scheduler.md
+++ b/docs/src/learning_rate_scheduler.md
@@ -65,8 +65,9 @@ Currently we will retain constant timestepping while we investigate further, tho
 
 Please let us know how you get on by setting the keyword argument in EKP
 ```julia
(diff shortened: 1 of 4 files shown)


eviatarbach:
LGTM! Thanks!

PR #295, opened 2023-06-08
Schedulers no longer break when particle failures occur

### LINKED ISSUE #292: Adaptive schedulers not compatible with particle failures
Currently schedulers meeting a NaN in the model evaluation will fail when calculating a timestep, even though we may be using a failure handler for the actual update. We should extend this behaviour to the schedulers too 

The only current case is (if we are not doing IgnoreFailures) to call the "split_by_success(g)" and pass only the the successful particles into the timestep calculation.

odunbar:
## Purpose 

Closes #292 

## Content

- added successful_ens filter to calculations of delta t
- added the `DataMisfitController()` into a test case which contains particle failures and it no longer breaks

----

### DIFF
--- a/src/LearningRateSchedulers.jl
+++ b/src/LearningRateSchedulers.jl
@@ -137,7 +137,7 @@ end
 """
 $(TYPEDSIGNATURES)
 
-Calculates next timestep by pushing to ekp.Δt,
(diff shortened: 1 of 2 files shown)


eviatarbach:
LGTM, thanks!

eviatarbach (approved):

PR #298, opened 2023-06-13
Add == for schedulers

### LINKED ISSUE #297: DataMisfitController() ==  DataMisfitController()  gives false
For some types, the `Scheduler() == Scheduler()` gives `false`, which is likely to cause test failures.
(Note that all the internal fields give `true` on "==")

Need to override `==` as done for `ParameterDistributions`  here:
https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/c746772a6fa8de37aefc15c249eb80d8c7ad854a/src/ParameterDistributions.jl#L557-L564

odunbar:
## Purpose 

Closes #297 

## Content

- add overload to `Base.(==)` for `LearningRateSchedulers`
- add tests

----

### DIFF
--- a/src/LearningRateSchedulers.jl
+++ b/src/LearningRateSchedulers.jl
@@ -308,3 +308,12 @@ function calculate_timestep!(
     end
     nothing
 end
+
+# overload ==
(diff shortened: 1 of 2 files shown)


eviatarbach (approved):
LGTM!

Issue #282, opened 2023-05-05
Gaussian Random Field parameters: Phase 2


odunbar:
With #252 merged, we have raised some new issues to accompany the GRF tools.

 - [ ] Check toml interface
 - [ ]  Document Darcy example
 - [ ]  Document GRF interface / parameter distribution
 - [ ]  compatability with new plot features

odunbar:
Deal with #300  here too

odunbar:
- [ ] As everything is stacked internally, we should provide an API to obtain the evaluation point grid, for easy reshaping (if we are plotting for example).

PR #304, opened 2023-06-29
Function parameter documentation

### LINKED ISSUE #300: `Base.show(::ParameterDistribution)` error with Gaussian Random Fields
To reproduce (from `ParameterDistributions` test ([link](https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/792ca73671ee187d1b556861e1eaee852546cb28/test/ParameterDistributions/runtests.jl#L308-L316)):
```julia
import GaussianRandomFields
const GRF = GaussianRandomFields
using EnsembleKalmanProcesses.ParameterDistributions
dim = 1
dofs = 10
pts = collect(0:0.01:1)
pkg = GRFJL()
grf = GRF.GaussianRandomField(GRF.CovarianceFunction(dim, GRF.Matern(0.05, 1)), GRF.KarhunenLoeve(dofs), pts)
d5 = GaussianRandomFieldInterface(grf, pkg)
c5 = bounded(-5, 5)
name5 = "grf_in_min5_5"
u5 = ParameterDistribution(d5, c5, name5)
```

gives this error:
```julia
julia> u5 = ParameterDistribution(d5, c5, name5)
Error showing value of type ParameterDistribution{GaussianRandomFieldInterface, Constraint{Bounded}, String}:
ERROR: BoundsError: attempt to access 1-element Vector{Constraint{Bounded}} at index [[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]]
Stacktrace:
   [1-3] ⋮ internal
       @ Base, Unknown
     [4] getindex(A::Vector{Constraint{Bounded}}, I::Vector{Int64})
       @ Base ./abstractarray.jl:1294
     [5] show(io::IOContext{Base.TTY}, distributions::ParameterDistribution{GaussianRandomFieldInterface, Constraint{Bounded}, …})
       @ EnsembleKalmanProcesses.ParameterDistributions ~/Documents/CliMA/EnsembleKalmanProcesses.jl/src/ParameterDistributions.jl:447
  [6-22] ⋮ internal
       @ Base.Multimedia, REPL, AbbreviatedStackTraces, REPL.LineEdit, Unknown
Use `err` to retrieve the full stack trace.

```

odunbar:
## Purpose 

Fixes #300 
Addresses GRF docs part of #282 

NB this is merely documenting what is currently implemented, improvements to the API will come soon.

## Content

- Adds the `constraint` keyword to fix `base.show()` bug with function distributions
- runs `show(parameter_distribution)` during tests to appease codecov
- Adds a section  and example of creating and plotting `GaussianRandomFieldInterface` object and samples.
Relevant docs section found [here](https://clima.github.io/EnsembleKalmanProcesses.jl/previews/PR304/parameter_distributions/#FunctionParameterDistributionType) and [here](https://clima.github.io/EnsembleKalmanProcesses.jl/previews/PR304/parameter_distributions/#function-example).

----

### DIFF
--- a/docs/src/parameter_distributions.md
+++ b/docs/src/parameter_distributions.md
@@ -8,7 +8,7 @@ Bayesian inference begins with an explicit prior distribution. This page describ
 
 A prior is specified by a `ParameterDistribution` object, which has three components:
 
- 1. The distribution itself, given as a [ParameterDistributionType](@ref) object. This includes standard Julia [Distributions](https://juliastats.org/Distributions.jl/stable/) as well as empirical/sample-based distributions, and can be uni- or multi-variate. To clarify, despite our use of the term "Kalman processes," the prior distribution is *not* required to be Gaussian.
+ 1. The distribution itself, given as a [ParameterDistributionType](@ref) object. This includes standard Julia [Distributions](https://juliastats.org/Distributions.jl/stable/), [GaussianRandomFields](https://github.com/PieterjanRobbe/GaussianRandomFields.jl) as well as empirical/sample-based distributions, and thus can be univariate, multivariate or functional. To clarify, despite our use of the term "Kalman processes," the prior distribution is *not* required to be Gaussian.
  2. A constraint (or array of constraints) on the domain of the distribution, given as a [ConstraintType](@ref) or `Array{ConstraintType}` object (the latter case builds a multivariate constraint as the Cartesian product of one-dimensional Constraints). This is used to enforce physical parameter values during inference: the model is never evaluated at parameter values outside the constrained region, and the posterior distribution will only be supported there.
  3. The parameter name, given as a `String`.
 
@@ -40,7 +40,6 @@ using EnsembleKalmanProcesses.ParameterDistributions # for `constrained_gaussian
 prior_1 = constrained_gaussian("param_1", μ_1, σ_1, lower_bound, upper_bound)
 prior_2 = constrained_gaussian("param_2", μ_2, σ_2, 0.0, Inf, repeats=3)
 prior = combine_distributions([prior_1, prior_2])
-nothing # hide
 ```
 
 `prior_1` is a `ParameterDistribution` describing a prior distribution for a parameter named `"param_1"` taking values on the interval [`lower_bound`, `upper_bound`]; the prior distribution has approximate mean ``μ_1`` and standard deviation ``σ_1``.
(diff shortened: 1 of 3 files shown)


docs/src/parameter_distributions.md
```diff
@@ -449,6 +463,51 @@ using Plots
...
 ```
 
+## [Function Distribution Example](@id function-example)
+Here, define a function parameter distribution on ``[0,1] \times [1,2]`` , bounded by ``[-5,-3]`` and with correlation lengthscales 0.05. First, we get the packages:
+
+```@example snip_fun
+using EnsembleKalmanProcesses.ParameterDistributions # For `ParameterDistribution`
+using Random, Distributions # for `rand` and `Normal`
+using Plots
+# IMPORT the GRF package
```
haakon-e, line 473: I suspect you're trying to say use `import` instead of `using` to load GRF. I think making that explicit could help readability.

docs/src/parameter_distributions.md
```diff
@@ -449,6 +463,51 @@ using Plots
...
+        "distribution" => grf,
+        "constraint" => bounded(-5, -3), 
+        "name" => "func_in_min5_min3",
+    )
+) # The ParameterDistribution with constraint in the output space
+nothing # hide
+```
+We plot 4 samples of this distribution. Samples are taken over the (30-dimensional) degrees of freedom, and then we apply the `transform_unconstrained_to_costrained` map to (i) build the function distribution, (ii) evaluate it on the numerical grid, and (iii) constrain the output with our prescribed bounds.
+```@example snip_fun
+shape = [length(pp) for pp in points]
```
haakon-e, line 505: could `ndims` be used here instead to illustrate the usage of that method?

haakon-e (approved):
Good PR! Added a few minor comments, but I basically think this can be merged. We can always make further improvements to the docs in the future as people start using the function parameters!

An aside:
It seems the terminology "class" or "subclass" is used a few locations throughout the documentation to describe parameter distributions both relating to function parameters, but also in general. I think it would be good to be a bit more precise to refer to these as abstract types and structs. That said, I think it's best to discuss and make those changes in a separate PR.

docs/src/parameter_distributions.md
odunbar: Done

docs/src/parameter_distributions.md
odunbar: the functional parameter `ndims` only gives the total number of eval points, where as here for plotting we need the breakdown of # eval points per dimension. It might be worth storing this grid-shape internally however, for future use in reshaping. I'll create a separate issue, I presume it is obtainable from the GRF package.

odunbar: I have appended this into #282

odunbar:
I've also created an issue for the terminology #308

PR #307, opened 2023-06-30
update example to the latest microphysics

### LINKED ISSUE #306: Compatibility issue while deploying new example to docs
# The Climate Modeling Alliance
### Purpose

The example file aerosol_activation.jl is only compatible with CloudMicrophysics v0.5, which constrains other CliMA packages to older versions. For example, CloudMicrophysics v0.5 is only compatible with the package SurfaceFluxes on versions up to 0.3. As a result, when adding an example file that depends on SurfaceFluxes v0.6, make.jl cannot build the docs for both aerosol_activation.jl and whatever file using SurfaceFluxes v0.6 due to compatibility mismatch. This poses a problem for branch wd/surface-flux-example, which intends to add the example kappa_calibration.jl, which relies on SurfaceFluxes v0.6.3. 

### Cost/Benefits/Risks

Although it takes time to update aerosol_activation.jl to match with more recent package releases, it makes the example file more relevant and easier to reproduce. In addition, by updating the file, it would allow developers such as myself to put out additional example files such as the one introduced above, kappa_calibration.jl, which depends on a newer version of SurfaceFluxes.

### People and Personnel

- Lead: @odunbar 
- Reviewers: @akshaysridhar, @LenkaNovak 

### Components

The solution would be updating aerosol_activation.jl to match current releases.

### Results and Deliverables

We would expect the updated example to have the same functionality but simply operate under the newer package releases.

trontrytel:
Hi. I have updated the example code to the latest CM.jl version. The example runs, but the convergence of the calibration is not as good as it used to be. 

I'll take another look at the microphysics code. But @odunbar could you also look at how we use the EKP package? Maybe something needs to be changed there as well?

Not-sure-if-closes #306

### DIFF
--- a/examples/AerosolActivation/Project.toml
+++ b/examples/AerosolActivation/Project.toml
@@ -10,4 +10,4 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Thermodynamics = "b60c26fb-14c3-4610-9d3e-2d17fe7ff00c"
 
 [compat]
-CloudMicrophysics = "0.5"
+CloudMicrophysics = "0.11"
--- a/examples/AerosolActivation/aerosol_activation.jl
+++ b/examples/AerosolActivation/aerosol_activation.jl
@@ -43,21 +43,24 @@ using Random
 rng_seed = 44
 rng = Random.seed!(Random.GLOBAL_RNG, rng_seed)
 
-using EnsembleKalmanProcesses
+import EnsembleKalmanProcesses as EKP
 using EnsembleKalmanProcesses.ParameterDistributions
-const EKP = EnsembleKalmanProcesses
 
-import CLIMAParameters
-const CP = CLIMAParameters
-struct EarthParameterSet <: CP.AbstractEarthParameterSet end
-const param_set = EarthParameterSet()
+import CLIMAParameters as CP
 
-import CloudMicrophysics
-const AM = CloudMicrophysics.AerosolModel
-const AA = CloudMicrophysics.AerosolActivation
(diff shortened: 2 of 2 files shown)


trontrytel:
Here are the old plots:

molar_mass_average_old
molar_mass_scatter_old
osmotic_coeff_average_old
osmotic_coeff_scatter_old

And here are the new:
[molar_mass_average.pdf](https://github.com/CliMA/EnsembleKalmanProcesses.jl/files/11915872/molar_mass_average.pdf)
[molar_mass_scatter.pdf](https://github.com/CliMA/EnsembleKalmanProcesses.jl/files/11915873/molar_mass_scatter.pdf)
[osmotic_coeff_average.pdf](https://github.com/CliMA/EnsembleKalmanProcesses.jl/files/11915874/osmotic_coeff_average.pdf)
[osmotic_coeff_scatter.pdf](https://github.com/CliMA/EnsembleKalmanProcesses.jl/files/11915875/osmotic_coeff_scatter.pdf)

odunbar:
I think that the EKP tools are working fine.

In terms of the experiment, i think the issue is simply that *the data is uninformative about the osmotic coefficient*. 

To express this, by shrinking the data noise standard deviation by a factor of 100 the molar mass converges always to the true value, while the osmotic coefficient varies and drifts around whatever the prior mean was set to e.g. here
molar_mass_average
osmotic_coeff_average

odunbar:
PS Here is one with the prior mean increased to 2.0
 
or with prior mean decreased to 0.1

odunbar:
@trontrytel any movement on this one?

PR #312, opened 2023-07-07
Removing unused `history` field in DataMisfitController

eviatarbach:
The `history` field in DataMisfitController no longer does anything. It should be removed.


odunbar (approved):
LGTM 

I think we were wondering about a redundancy of information storage, but happy to remove until needed!

PR #320, opened 2023-08-05
Ensemble transform Kalman inversion

### LINKED ISSUE #299: Time complexity of EKI
Probably won't be an issue in the near future, but I believe that EKI as currently implemented is O(p^3), where p is the observation dimension. However, based on the idea of [ensemble square-root Kalman filters](http://doi.org/10.1175/1520-0493(2003)131<1485:ESRF>2.0.CO;2), this can be reduced to O(p). I believe that some of the EKI variants implemented in [Huang et al. (2022)](http://doi.org/10.1088/1361-6420/ac99fa) have this property.

### LINKED ISSUE #324: Implement and test ensemble transform Kalman inversion (ETKI)

eviatarbach:
Closes #324


eviatarbach:
The initial implementation seems to be working, but still needs tests, documentation, and examples. Note that the ETKI does not work with localization.

odunbar:
Any developments here? It would be good maybe to just have the source code and unit tests in this PR, we can make another for examples & docs if you like

src/EnsembleTransformKalmanInversion.jl
```diff
@@ -0,0 +1,137 @@
+#Ensemble Transform Kalman Inversion: specific structures and function definitions
+
+"""
+    TransformInversion <: Process
+
```
odunbar, line 5: Add
```# Fields

$(TYPEDFIELDS)
```
and put the `# Inverse ... matrix` in quotes instead of a comment `" " `

src/EnsembleTransformKalmanInversion.jl
```diff
@@ -0,0 +1,137 @@
...
+    u::AbstractMatrix{FT},
+    g::AbstractMatrix{FT},
+    y::AbstractVector{FT},
+    obs_noise_cov::Union{AbstractMatrix{CT}, UniformScaling{CT}},
+) where {FT <: Real, IT, CT <: Real}
+
+    m = size(u, 2)
+    Γ_inv = ekp.process.Γ_inv
+    # N_obs × N_obs \ [N_obs × N_ens]
+    # --> tmp is [N_obs × N_ens]
```
odunbar, line 61: I assume this comment is an artifact from somewhere? as you don't create tmp here

test/EnsembleKalmanProcess/runtests.jl
```diff
@@ -607,6 +607,108 @@ end
...
+
+    # Seed for pseudo-random number generator
+    rng = Random.MersenneTwister(rng_seed)
+
+    initial_ensemble = EKP.construct_initial_ensemble(rng, prior, N_ens)
+
+    ekiobj = nothing
+    eki_final_result = nothing
+
+    for ((i_prob, inv_problem), eks_final_result, eksobj) in zip(enumerate(inv_problems), eks_final_results, eksobjs)
```
odunbar, line 620: Are the eks objects used anywhere?

odunbar (commented):
Hey! Thanks a lot for this! 
- In what you have implemented I only tiny comments here - it all looks good otherwise!

One other comment, could we have a test where we perform one update with an increasingly high-dimensional output sized output space? e.g. `10^2, 10^3, 10^4, 10^5`  and time it? You could either just println the times, or you could compare loosely to theoretical scalings. 

It would be good to have something like this to warn us in case future features corrupt the scalings for this method.

Once this is implemented, could you add the times to the PR description and I will be happy for merge!

Issue #321, opened 2023-08-09
Momentum-accelerated EKP


odunbar:
Offline we have been experimenting with a few momentum-based accelerations of EKI with good success. Integration of such approaches into the code for application with general EKP-type methods is detailed below. In general we have identified a rough template of what such accelartors will look like:

```julia 
# type for flow accelerators
abstract type Accelerator

# e.g. the Nesterov type contains the parameter and a history of states
Nesterov <: Accelerator 
    momentum_parameter
    state_history
end

# it lives in the general EKP object
struct EnsembleKalmanProcess{..., A<:Accelerator}
    ...
    accelerator::A
end

# the state history is intialized with the initial ensemble
function initial_ensemble(...)
    ...
    ekp.accelerator.state_history = initial_ensemble 
end

# the update involves updating the particles with the usual method, along with the augmented state
function update_ensemble(u,g)
    update_ensemble!(u,g,ekp.process)  # update to perform an approximate gradient step
    new_ensemble = accelerator_update(ekp.accelerator, u[end]) # shift new ensemble by accelerator
    ekp.accelerator.state_history = new_ensemble
    ...
return new_ensemble
end
```
It is also important that the saved pairs are the momentum-shifted ensemble members and the evaluations of them (i.e. `v` and `G(v)` remain consistent). This works because accelerators typically still involve the calculation of a typical gradient-based update, and are simply followed by a perturbation of the particles by momentum. Hopefully UKI,EKI,Transform-EKI,SEKI, (maybe even EKS which also represents a gradient flow) can be all be updated with the same interface.

PR #322, opened 2023-08-15
WIP: adding momentum-inspired accelerators to EKP

### LINKED ISSUE #323: Implement Source code and Unit tests in a method-independent way

sydneyvernon:
## Purpose 

Closes #323 

## Content

----
## Current results from nonlinear inversion unit tests:

```
Accelerator: DefaultAccelerator Process: Inversion
┌ Info: Convergence:
│   cost_initial = 55.61433558387613
└   cost_final = 8.644475010222061
Accelerator: NesterovAccelerator Process: Inversion
┌ Info: Convergence:
│   cost_initial = 55.32972049471293
└   cost_final = 12.14289870253091
Accelerator: DefaultAccelerator Process: Sampler
┌ Info: Convergence:
│   cost_initial = 56.12630362217311
└   cost_final = 13.797245580534
Accelerator: NesterovAccelerator Process: Sampler
┌ Info: Convergence:
│   cost_initial = 58.07591023338068
└   cost_final = 7.951552079718973
Accelerator: DefaultAccelerator Process: TransformInversion
┌ Info: Convergence:
│   cost_initial = 58.242226993212626
└   cost_final = 5.25456484944421
Accelerator: NesterovAccelerator Process: TransformInversion
┌ Info: Convergence:
│   cost_initial = 58.64590869107448
└   cost_final = 4.183960610134597
```


src/Accelerators.jl
```diff
@@ -0,0 +1,47 @@
...
+"""
+$(TYPEDEF)
+
+Default accelerator provides no acceleration.
+
+$(TYPEDFIELDS)
+"""
+struct DefaultAccelerator <: Accelerator end # can use types/subtypes to ensure v, ens_prev consistent, and to multiple dispatch if multiple cases.
+
+function DefaultAccelerator()
```
odunbar, line 16: You should remove this constructor - as (1) The `DefaultAccelerator()` constructor is automatically created when defining the struct above, and (2) we do not want this to return the Nesterov Accelerator.

src/Accelerators.jl
```diff
@@ -0,0 +1,47 @@
...
+$(TYPEDFIELDS)
+"""
+struct NesterovAccelerator{FT<:AbstractFloat, VV<:AbstractVector} <: Accelerator  ## extra safety to ensure types of v,ens_prev match?
+    r::FT
+    v
+    ens_historical::VV
+end
+
+function NesterovAccelerator()
+    return NesterovAccelerator
```
odunbar, line 35: Because the struct you create above has 3 entries, you need this to return some default version e.g. 
```julia

return NesterovAccelerator(3,Float64[],Float64[])
```
So when the user calls `NesterovAccelerator()`, it will return the version with `r=3`

src/SparseEnsembleKalmanInversion.jl
```diff
@@ -222,12 +222,13 @@ function update_ensemble!(
...
 
     # store new parameters (and model outputs)
-    push!(ekp.u, DataContainer(u, data_are_columns = true))
     push!(ekp.g, DataContainer(g, data_are_columns = true))
 
     # Store error
     compute_error!(ekp)
 
     # Check convergence
     cov_new = cov(get_u_final(ekp), dims = 2)
```
odunbar, line 231: Don't forget that lines like this need to become:
```julia
cov_new = cov(u, dims = 2)
```

odunbar: in EKS too

odunbar (commented):
Looks great so far - just adding some hints as you go. 

As you may have seen the Unscented version is a bit trickier to parse, so we can look at how to get an interface working here later.

src/Accelerators.jl
sydneyvernon: So when we define the struct, there is a constructor automatically made for an object literally called "DefaultAccelerator"?

sydneyvernon:
Last commit shows an initial version (not yet functional) of what NesterovAccelerator could look like. Notes:
- **incorporating additional states**: I am trying to be as noninvasive as possible and keep my contributions contained in `Accelerators.jl`. In the current framework, I think I would have to change existing (widely used) functions `get_u` or `update_ensemble` if I wanted to implement momentum EKI exactly as we derived it, because momentum EKI follows 2 states (_v_ and _u_), and evaluates G on _v_ but returns _u_ as the ensemble value. However, I don't see any problem with accepting _v_ as the "ensemble value" and storing information on _u_ for acceleration purposes only. I looked at the convergence of the _v_ state in my smaller project/examples, and it seems to work just fine (?). Only issue is that I will have to be mindful of potentially shifting my iteration index.
- **timestep**: again hoping to keep things contained to `Accelerators.jl`: EKI and momentum EKI apply virtually the same update, but scale by _dt_ and _dt^2_ respectively. For now I haven't addressed this since for individual applications, if we have optimized timesteppers in place, I don't think this is significant(?). I think the issue may come when comparing methods, though.

odunbar:
First, if you have addressed my comments above, feel free to "Resolve conversation" on them

On your points
- (1) is indeed what is described in #321, that is, the ensemble update produces `new_ens` (the old "stored state") but this  now gets moved by the accelerator before being stored as `ekp.u`. So following this structure should produce the desired result
- (2) excellent point about the timestep. I think it would be good to be consistent, but I will add this to the issue, rather than dealing with it in this PR.

eviatarbach:
Yes, good point about the timestep. Note also that the timestep is implemented in EKP as a rescaling of the observation error covariance; please check the consistency of this with the timestep in momentum EKI.

test/runtests.jl
```diff
@@ -25,15 +25,15 @@ end
     end
 
     for submodule in [
```
odunbar, line 27: Note: this file shouldn't be added, it removes a lot of our online testing.

src/EnsembleKalmanProcess.jl
```diff
@@ -189,23 +196,32 @@ function EnsembleKalmanProcess(
     # timestep store
     Δt = FT[]
 
+    # set accelerator ## will change to allow for additional accelerators in future
+    if isnothing(accelerator)
+        acc = DefaultAccelerator()
+    else
+        acc = NesterovAccelerator(accelerator.r, params)
```
eviatarbach, line 203: This is kind of awkward, since you are recreating the `Accelerator` struct. Can you use the `update_state` function here instead?

src/Accelerators.jl
```diff
@@ -0,0 +1,65 @@
...
+
+$(TYPEDFIELDS)
+"""
+mutable struct NesterovAccelerator{FT<:AbstractFloat} <: Accelerator
+    r::FT
+    u_prev::Any
+end
+
+function NesterovAccelerator()
+    return NesterovAccelerator(3.0,Float64[])
```
eviatarbach, line 25: Please make `r` a parameter to the constructor, but you can set a default of 3.

odunbar:
Some notes regarding UKI. It saves only the mean and cov of the ensemble, in `process.u_mean` and `process.uu_cov`
A possible strategy (I'd be happy to implement if you are tight on time etc.)
- Make these returned, not saved, (as we did with other method ensembles). So for example UKI will return u_mean, uu_cov, and also the sigma points.
- in the current `update_state!(ekp,u)` we define a new update method for UKI typed `ekp`. This shifts the mean, and contracts the covariance appropriately as with the acceleration. It then saves them in the process object, and redraws the new ensemble ("sigma points") and saves these.

src/Accelerators.jl
```diff
@@ -0,0 +1,90 @@
...
+"""
+Placeholder state update method for UKI with Nesterov Accelerator.
+The Accelerator framework has not yet been integrated with UKI process, so this
+method throws an error.
+"""
+function update_state!(
+    ekp::EnsembleKalmanProcess{FT, IT, P, LRS, NesterovAccelerator},
+    u::MA,
+) where {FT <: AbstractFloat, IT <: Int, P <: Unscented, LRS <: LearningRateScheduler, MA <: AbstractMatrix{FT}}
+    throw(Error("Acceleration is not yet implemented for UKI."))
```
odunbar, line 89: Could you change this to
`throw(ArgumentError("option `accelerator = NesterovAccelerator`  is not implemented for UKI, please use `DefaultAccelerator`"))`

You can test for codecov it with something like this
`@test_throws ArgumentError update_state!(uki(...,accelerator = NesterovAccelerator), zeros(2,2))`

src/EnsembleKalmanProcess.jl
```diff
@@ -193,23 +206,36 @@ function EnsembleKalmanProcess(
...
+    end
+
     # failure handler
     fh = FailureHandler(process, failure_handler_method)
     # localizer
     loc = Localizer(localization_method, N_par, N_obs, N_ens, FT)
 
     if verbose
-        @info "Initializing ensemble Kalman process of type $(nameof(typeof(process)))\nNumber of ensemble members: $(N_ens)\nLocalization: $(nameof(typeof(localization_method)))\nFailure handler: $(nameof(typeof(failure_handler_method)))\nScheduler: $(nameof(typeof(lrs)))"
+        @info "Initializing ensemble Kalman process of type $(nameof(typeof(process)))\nNumber of ensemble members: $(N_ens)\nLocalization: $(nameof(typeof(localization_method)))\nFailure handler: $(nameof(typeof(failure_handler_method)))\nScheduler: $(nameof(typeof(lrs)))\nAccelerator: $(nameof(typeof(acc)))"
```
odunbar, line 227: I would ignore testing this as it is a print statement (just run it once locally with the flag verbose=true to make sure it runs

src/EnsembleKalmanProcess.jl
```diff
@@ -423,6 +448,14 @@ function get_scheduler(ekp::EnsembleKalmanProcess)
     return ekp.scheduler
 end
 
+"""
+    get_accelerator(ekp::EnsembleKalmanProcess)
+Return accelerator type of EnsembleKalmanProcess.
+"""
+function get_accelerator(ekp::EnsembleKalmanProcess)
```
odunbar, line 455: It is good this was picked up by codecov
I think if you try to test it you will get an error that it doesnt exist because I don't think we export it,

You should go to the top of the file and add `export get_accelerator` and write a short test that calls it

odunbar (commented):
Hi Sydney a couple of comments to help us to the finish line on this one

src/EnsembleKalmanInversion.jl
```diff
@@ -49,7 +49,7 @@ from eqns. (4) and (5) of Schillings and Stuart (2017).
 Localization is implemented following the `ekp.localizer`.
 """
 function eki_update(
-    ekp::EnsembleKalmanProcess{FT, IT, Inversion},
+    ekp::EnsembleKalmanProcess{FT, IT, Inversion}, ## where is LRS type here (so i know where to put acc type?)
```
odunbar, line 52: remove this comment?

test/EnsembleKalmanProcess/runtests.jl
```diff
@@ -79,6 +79,123 @@ inv_problems = [inv_problems..., nl_inv_problems...]
...
+        final_means = []
+
+        for (accelerator, N_iter) in zip(accelerators, N_iters)
+            println("Accelerator: ", nameof(typeof(accelerator)), " Process: ", nameof(typeof(process)))
+            if !(nameof(typeof(process)) == Symbol(Unscented))
+                ekpobj = EKP.EnsembleKalmanProcess(
+                    initial_ensemble,
+                    y_obs,
+                    Γy,
+                    process,
```
odunbar, line 158: To be safe, I would create a new process object for each test. There is no guarantee that they aren't modified during the calls to EKP updates.

test/EnsembleKalmanProcess/runtests.jl
```diff
@@ -79,6 +79,123 @@ inv_problems = [inv_problems..., nl_inv_problems...]
...
+                    rng = copy(rng),
+                    accelerator = accelerator,
+                )
+            end
+            for i in 1:N_iter
+                params_i = get_ϕ_final(prior, ekpobj)
+                g_ens = G(params_i)
+                if i == 3
```
odunbar, line 166: This i==3 statement was for testing convergence in an earlier test and can be removed

test/EnsembleKalmanProcess/runtests.jl
```diff
@@ -79,6 +79,123 @@ inv_problems = [inv_problems..., nl_inv_problems...]
...
+        Inversion(),
+        scheduler = DefaultScheduler(),
+        accelerator = DefaultAccelerator(),
+    )
+    eksobj_noacc_specified = EKP.EnsembleKalmanProcess(
+        initial_ensemble,
+        y_obs,
+        Γy,
+        Sampler(prior),
+        scheduler = DefaultScheduler(),
```
odunbar, line 121: I would remove all the scheduler = DefaultScheduler() comments, and use the default for each method (this will remove the EKS warnings)

odunbar (approved):
LGTM! Thanks for all the work!

eviatarbach:
This looks great Sydney! Thank you for all the hard work!

It almost looks ready to go. Could you please just verify that `TransformInversion` works with acceleration, and add a test for it?

odunbar:
So we did add a unit test for ETKI (see L116 in the added runtests file) which seems to run and converge, so i think we are covered there. I think we can defer actual acceleration for the next PR (with proper examples).

For EKS, this is a good point. I think there are two options, either a warning on creation of the accelerator. Or we just put something in the docs. I think perhaps you are correct and we should do both. 

Sydney, could you add a warning for when someone creates the accelerator and the EKP object is of type Sampler? Just saying that it is an experimental feature and may affect behaviour

eviatarbach:
LGTM!

PR #326, opened 2023-09-18
Bugfix darcy plot

### LINKED ISSUE #325: darcy.jl: problem when plotting true flow
In the plotting code in `examples\Darcy\darcy.jl`, we plot "true flow" in line 169 using variable `h_2d` which was mistakenly overwritten on line 155 and no longer represents the true flow. 
This can be quickly fixed by either (1) re-evaluating darcy model on `κ_true` before plotting true flow or (2) renaming `h_2d` in lines 80 and 81 as `h_2d_true`

odunbar:
## Purpose 

Closes #325 

## Content

- variable rename
- [ALSO] docs build error fixed (due to updated Documenter version) by removing `strict=true`

----


odunbar:
Merge on test-pass, bugfix is simply a variable name substitution

PR #348, opened 2023-11-03
Updated accelerators with variable timestep, add localization case to examples

odunbar:
## Purpose 

- removal of explicit timestep dependency of `NesterovAccelerator()` appears to be a more robust approach with a do-no-harm adjustment

## Content

- removal of `Delta t` and `Delta t_prev` dependency
- add fig_save directories and fig_save naming based on timestep for accelerated examples. 
- add some info statements, more encapsulated "cases", (i) constant stepper, (ii) dmc stepper (iii) dmc with localization and small ensemble (size = 5)
- Add GC into Darcy example over random trials.
- Added simple `ConstantNesterovAccelerator` which sets the momentum parameter = constant
- Renamed `ConstantStepNesterovAccelerator` to `FirstOrderNesterovAccelerator`
- Fixed an off-by-one error in `FirstOrder`

### Example figures from `Accelerated/`
- **KEY INSIGHT:** Acceleration still proves to have increased benefit for small ensembles while using localization (common in practical scientific needs)

`ExpSinAccelerated` for `case = const` , `case = dmc` and `case = dmc-loc-small-ens`:

`LorenzAccelerated` for `case = const` , `case = dmc` and `case = dmc-loc-small-ens`:

`DarcyAccelerated` for `case = const` , `case = dmc` and `case = dmc-loc-small-ens`:

----

### DIFF
--- a/examples/Accelerators/DarcyAccelerated/darcy_accelerated.jl
+++ b/examples/Accelerators/DarcyAccelerated/darcy_accelerated.jl
@@ -169,7 +169,7 @@ function main()
             err_acc_cs[i] = log.(get_error(ekiobj_acc_cs)[end])
             errs_acc_cs[trial, :] = err_acc_cs
         end
-        
+
         GC.gc()
 
     end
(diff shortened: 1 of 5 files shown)


eviatarbach (approved):
LGTM!

PR #353, opened 2023-11-09
Add `constrained_gaussian` TOML parser

nefrathenrici:
A few notes:
- The test for this parser directly compares to a `Parameterized(Normal())` distribution rather than the `constrained_gaussian` constructor
- If the distribution being parsed is a constrained gaussian, the constraint constructor is bypassed

### DIFF
--- a/src/TOMLInterface.jl
+++ b/src/TOMLInterface.jl
@@ -5,7 +5,6 @@ using TOML
 using Distributions
 using EnsembleKalmanProcesses.ParameterDistributions
 
-
 # Exports
 export path_to_ensemble_member
 export get_parameter_distribution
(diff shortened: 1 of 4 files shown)


src/TOMLInterface.jl
```diff
@@ -243,6 +247,13 @@ function get_distribution_from_expr(d::Expr)
 
         return dist_type(dist_args)
 
+    elseif dist_type_symb == Symbol("constrained_gaussian")
```
odunbar, line 250: There can also be a `repeats=N` optional argument that should be passed through too,

nefrathenrici: Do we need full kwarg functionality or just `repeats=N`?

src/TOMLInterface.jl
```diff
@@ -243,6 +246,30 @@ function get_distribution_from_expr(d::Expr)
...
+            repeats = parse_repeats(d.args[2].args)
+            index += 1
+            # Positional kwargs
+        elseif length(d.args) > 6 && d.args[7].head == :kw
+            repeats = parse_repeats([d.args[7]])
+        end
+        name, dist_mean, dist_std, lb, ub = d.args[index:(index + 4)]
+        name = string(name)
+        lower_bound, upper_bound = parse.(Float64, string.((lb, ub)))
+        return constrained_gaussian(name, dist_mean, dist_std, lower_bound, upper_bound; repeats)
```
odunbar, line 272: could you pass repeats as `repeats = repeats` still?

odunbar (approved):
LGTM after 
- you address the small comment, and 
- you add a `test_throws` argument for the  wrong repeats type?

PR #360, opened 2023-11-18
remove positive definiteness constraints, allow user defined additive inflation

### LINKED ISSUE #359: Positive definite corrections in `get_u_cov`
https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/0db3339649cb29a55933d7eb4e711e6236c9fce0/src/EnsembleKalmanProcess.jl#L612
I often find that the additive inflation breaks due to non-positive definiteness when estimating covariances, also the `get_u_cov_final()` may not break the subspace property.

We can resolve both of these issues by adding either
1. the function `posdef_correct()` around the `get_u_cov_XXX(ekp)` when estimating the covariances, 
2. the function `posdef_correct()` within the definition of `get_u_cov()` and `get_u_cov_prior` themselves. (so this change would apply more broadly)

I prefer option 2, as i can't see why you wouldn't want positive definiteness. Though it would slow down the call to this function. However, `get_u_cov` is only called within the source code in the additive inflation step, and once per iteration for EKS & UKI. The function is also occasionally used by users to obtain diagnostics. 

NB posdef_correct currently lives in the `LearningRateSchedulers` file for some reason

odunbar:
## Purpose 

Closes #359

## Content

- Change all `MvNormal(m,C)` to `m .+ sqrt(C) * MvNormal(0,I)`, which can handle semi-definite "C"
- bugfix: Add `rng` into sample_empirical_gaussian
- Remove DMC test for Sparse inversion as it no longer seems to converge. _I believe this was likely unstable even in previous version of the code, stick with constant timesteppers_
-  allow users to provide an additive inflation matrix, defaults to prior covariance. `additive_inflation_cov = ...` in update to provide this. This allows for users to break the subspace property
- Unit tests for the new interface
- updated the inflation docs page [here](https://clima.github.io/EnsembleKalmanProcesses.jl/previews/PR360/inflation/)

### Testing 
E.g in `Examples/LorenzAccelerated/lorenz_accelerated.jl`,  on `main` the following breaks
```julia
update_ensemble!(...,additive_inflation=true, s=1e-2)
``` 
while it works on the current branch (with different choices of inflation matrix too)
```julia
update_ensemble!(...,additive_inflation=true, s=1e-2)
X = 0.1*I
update_ensemble!(...,additive_inflation=true, additive_inflation_cov = X, s=1e-2)
```
## Interface Note
Before the inflation default was effectively to inflate using `additive_inflation_cov = get_cov_u_final(ekp)`  now it defaults to the  `additive_inflation_cov = get_cov_u_prior(ekp)`. This is because you are likely not going to use the former situation - this is better completed using multiplicative inflation.

----

### DIFF
--- a/docs/src/inflation.md
+++ b/docs/src/inflation.md
@@ -33,28 +33,25 @@ Multiplicative inflation can be used by flagging the `update_ensemble!` method a
 ```
 
 ## Additive Inflation 
-Additive inflation is implemented by systematically adding stochastic perturbations to the parameter ensemble in the form of Gaussian noise. Additive inflation breaks the linear subspace property, meaning the parameter ensemble can evolve outside of the span of the initial ensemble. In additive inflation, the ensemble is perturbed in the following manner after the standard Kalman update:
+Additive inflation is implemented by systematically adding stochastic perturbations to the parameter ensemble in the form of Gaussian noise. Additive inflation is capable of breaking the linear subspace property, meaning the parameter ensemble can evolve outside of the span of the current ensemble. In additive inflation, the ensemble is perturbed in the following manner after the standard Kalman update:
 
 ```math
      u_{n+1} = u_n + \zeta_{n} \qquad (3) \\
-    \zeta_{n} \sim N(0, \frac{s \Delta{t} }{1 - s \Delta{t}} C_n) \qquad (4)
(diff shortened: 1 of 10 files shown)


eviatarbach (approved):
LGTM! Thank you for working on this.

PR #368, opened 2024-02-22
Add troubleshooting doc

costachris:
Add troubleshooting page to docs detailing how to deal with common practical issues.


haakon-e:
@costachris Thanks for this, looks really useful!

I think something I was thinking of when I initially brought this up was something along the lines of the PyDMD docs (['Current PyDMD Capabilities"](https://github.com/PyDMD/PyDMD?tab=readme-ov-file#description)) and (["Which method should I use"](https://github.com/PyDMD/PyDMD?tab=readme-ov-file#using-pydmd)):
image
and
image


For the latter, I've drafted a flow chart for this repo [[link](https://miro.com/app/board/uXjVNm_1teY=/?share_link_id=666335413970)]:
image

odunbar:
I like these ideas - some thoughts on the general scope
- I like the idea of the 2 plots. One to highlight what exists, and one to highlight what users should pick.
- The platform will require the occasional update as new features etc are added. We should think about the best way to do this? Miro is probably ok, but I think anyone can change the information at the Miro link, so this may need to be private in some way?

RE: the implementation you suggest, looks good but it feels like there are several flows that end up overlapping (someone could look at 3 of them and be told to use 3 different methods)
Perhaps the division in these 4 flows can be separated, so that they don't often . Titles for these could be based on the different stages that the user asks, such as
1. Which method should I choose? 
2. Which additional features might I need? 
3. Which priors should I choose? 
4. Is the method working?

The initial starts of the flow could be e.g. 
1. My goal is to learn parameters... (sparsely, from huge amounts of data, with uncertainty...)
2. I'm constrained by... (running only small ensembles, a quite unstable model)
3. I know my parameter priors are... (Have a known distribution, only have known constraints, are Vectors/functions, coming from a histogram)
4. I see that ... (what you have is good for this, I would recommend also adding a "verbose" flag)

I think there are likely better implementations that what I suggest, but maybe this helps disentangle

docs/src/troubleshooting.md
```diff
@@ -0,0 +1,21 @@
+# Troubleshooting and Workflow Tips
+
+## High Failure Rate
+
+While some EKI variants include failure handlers, excessively high failure rates (i.e., > 80%) can lead to inversions finding local minima or failing to converge. To address this:
+
+- **Stabilize the Forward Model**: Ensure the forward model remains stable for small parameter perturbations in offline tests.
```
odunbar, line 7: maybe add: "If the forward model is unstable for most of the parameter space, it is challenging to explore it".

odunbar, line 7: maybe add: "If the forward model is unstable for most of the parameter space, it is challenging to explore it with a calibration method".

docs/src/troubleshooting.md
```diff
@@ -0,0 +1,21 @@
...
+- **Adjust Priors**: Reduce the uncertainty in priors. Priors with large variances can lead to forward evaluations that deviate significantly from the known prior means, increasing the likelihood of failures.
+- **Increase Ensemble Size**: Without localization (or other methods that break the subspace property), the ensemble size should generally exceed the number of parameters being optimized. The ensemble size needs to be large enough to ensure a sufficient number of successful runs, given the failure rate.
+- **Increase Preconditioner Retry Count**: If using a preconditioner that recursively draws from the prior until a stable run is achieved, increase the retry count (typically to more than 5).
```
odunbar, line 10: Is this something in EKP? If not maybe you are _suggesting_ use of a preconditioner that recursively draws from the prior. Note this is inefficient

docs/src/troubleshooting.md
```diff
@@ -0,0 +1,21 @@
...
+- **Implement Parameter Inflation**: Prevent the ensemble from collapsing prematurely by adding parameter inflation.
```
odunbar, line 11: - premature collapse is a consequence of the failure handling
- link to inflation docs page

docs/src/troubleshooting.md
```diff
@@ -0,0 +1,21 @@
+# Troubleshooting and Workflow Tips
+
+## High Failure Rate
+
+While some EKI variants include failure handlers, excessively high failure rates (i.e., > 80%) can lead to inversions finding local minima or failing to converge. To address this:
+
+- **Stabilize the Forward Model**: Ensure the forward model remains stable for small parameter perturbations in offline tests.
+- **Adjust Priors**: Reduce the uncertainty in priors. Priors with large variances can lead to forward evaluations that deviate significantly from the known prior means, increasing the likelihood of failures.
+- **Increase Ensemble Size**: Without localization (or other methods that break the subspace property), the ensemble size should generally exceed the number of parameters being optimized. The ensemble size needs to be large enough to ensure a sufficient number of successful runs, given the failure rate.
```
odunbar, line 9: link to localizer page?

docs/src/troubleshooting.md
```diff
@@ -0,0 +1,21 @@
...
+- **Increase Preconditioner Retry Count**: If using a preconditioner that recursively draws from the prior until a stable run is achieved, increase the retry count (typically to more than 5).
+- **Implement Parameter Inflation**: Prevent the ensemble from collapsing prematurely by adding parameter inflation.
+
+## Loss Doesn't Converge
+
+If the loss decreases too slowly or diverges:
+
+- **Check for Failures**: Refer to the suggestions for handling a high failure rate.
+- **Adjust the Artificial Timestep**: For indirect learning problems involving neural networks, larger timesteps [O(10)] are generally more effective and using variable timesteppers (e.g., DMC) tends to yield the best results.
```
odunbar, line 18: -call it `DataMisfitController()`
-link to scheduler page?

docs/src/troubleshooting.md
```diff
@@ -0,0 +1,21 @@
...
+- **If Batching, Increase Batch Size**: If the loss is too noisy and convergence is slow, consider increasing the batch size.
```
odunbar, line 19: link to page where we talk about batches? I'm not sure it is mentioned much in the docs - inflation page?

docs/src/troubleshooting.md
```diff
@@ -0,0 +1,21 @@
...
+- **Check Observation Noise in Data Space**: Ensure that noise estimates are realistic and consistent across variables with different dimensions and variability characteristics. Observation noise that is unrealistically large for a given variable or data point may prevent convergence to solutions that closely fit the data. Carefully base noise estimates on empirical data or domain knowledge, and try reducing noise if the previous suggestions don’t work.
```
odunbar, line 20: This is important, perhaps add that this is VERY common if the user is just using `\sigma^2 * I` as artificial noise

odunbar: Perhaps could also rephrase to "Even if `u` does not look right, look at graphs of `G(u)` compared to `y +/- 2*sigma` " to see if the forward map is within the noise level". If it is then one cannot converge "further" without smaller noise / different loss. If it is not then this may indicate structural model errors between model and data leading to parameter trade-offs

odunbar (commented):
Hi @costachris, 

Sorry I did not review for a while, I did not see this page - I was looking at the flowcharts only. looking good! 
- I added a couple of notes where we should link to other docs pages with like `[...](@ref ...)` 
- some features like batching, preconditioning, are not base-EKI features. So need a bit more detail/clarity

docs/src/troubleshooting.md
costachris: it's not natively in EKP. will modify the description to reflect that.

costachris:
@odunbar , we went through your comments and addressed them. Let us know if we missed anything or if you have anymore suggestions.

docs/src/troubleshooting.md
```diff
@@ -0,0 +1,21 @@
+# Troubleshooting and Workflow Tips
+
+## High Failure Rate
+
+While some EKI variants include failure handlers, excessively high failure rates (i.e., > 80%) can lead to inversions finding local minima or failing to converge. To address this:
+
+- **Stabilize the Forward Model**: Ensure the forward model remains stable for small parameter perturbations in offline tests. If the forward model is unstable for most of the parameter space, it is challenging to explore it with a calibration method.
+- **Adjust Priors**: Reduce the uncertainty in priors. Priors with large variances can lead to forward evaluations that deviate significantly from the known prior means, increasing the likelihood of failures.
+- **Increase Ensemble Size**: Without localization, or other methods that break the subspace property (see [localization](@ref localization) for more), the ensemble size should generally exceed the number of parameters being optimized. The ensemble size needs to be large enough to ensure a sufficient number of successful runs, given the failure rate.
```
odunbar, line 9: could just link the localization in `without localization...`? rather than adding the parenthesis

docs/src/troubleshooting.md
```diff
@@ -0,0 +1,21 @@
...
+- **Check Observation Noise in Data Space**: Ensure that noise estimates are realistic and consistent across variables with different dimensions and variability characteristics. Observation noise that is unrealistically large for a given variable or data point may prevent convergence to solutions that closely fit the data. Carefully base noise estimates on empirical data or domain knowledge, and try reducing noise if the previous suggestions don’t work. This is especially common if using ``\sigma^2 * I`` as artificial noise. Even if ``u`` appears incorrect, it is advisable to examine the graphs of ``G(u)`` compared to  ``y \pm 2\sigma`` to determine if the forward map lies within the noise level. If it does, further convergence cannot be achieved without reducing the noise or altering the loss function. If it does not, this discrepancy might suggest inherent structural errors between the model and the data, which could lead to trade-offs in parameter estimation.
```
odunbar, line 20: The point on "Check obseration noise in observation space" is more about finding even if it has converged. Maybe we  could put this in a different section, or highlight it as a first-thing-to-do

odunbar (approved):
Thanks - it looks a lot better, 

Just a couple of small points left and then I am happy for merge

README.md
```diff
@@ -36,3 +36,5 @@ If you use the examples or code, please cite our article at JOSS in your publish
 ### Requirements
 Julia version 1.6+ 
 
+### Getting Started 
+![eki-getting-started](https://github.com/CliMA/EnsembleKalmanProcesses.jl/assets/45243236/5baf2de9-d55e-41f4-bda9-82f416900bb3)
```
odunbar, line 40: Review of figure:
General note: 
- The language in different grey boxes is a bit inconsistent (
- Capitalization/Punctuation is inconsistent
- Maybe we should emphasize using the (more searchable) names we have in the code e.g. `TransformInversion` rather than `ETKI` or `Localizers` rather than `localization`

"Which Methods..."
- "Simply" -> Use EKI: `Inversion()`
- did you wish to include the code-name i.e Sparse EKI is `SparseInversion`, UKI is `Unscented`, ETKI is `TransformInversion`
- ETKI is for high-dimensional data
- I would remove the "Note: " as these are typically included in the description in the tool on the docs, and the Note's that are there are currently inconsistent between tools
- It's weird having CES branching with arrows from EKS/UKI - though I'm not too fussed (maybe just have a dotted line, or have a disjointed box?

Additional features:
- "Try using SampleSuccGauss" -> "Try using our FailureHandler SampleSuccGauss"
- "if you cannot..." box just replace with "Try using our Localizers"

Which Priors should I choose:
- Maybe just make the grey boxes read a little more consistently
- "only have known constraints" - remove word "only"
- We should impress that we want people to use the constrained_gaussian constructor in most cases, as it's easy to use!

Is the method working?
- say we have a troubleshooting page
- parallelization is spelled incorrectly

odunbar (commented):
Hi @haakon-e 
I have a lengthy comment on the diagram, I like the structure and it looks good! I think mainly it's just language consistency, and leaning towards using the searchable code terms rather than acronyms perhaps?

README.md
haakon-e: Thanks for the feedback and discussion. Updated figure is now in the branch.

odunbar (approved):
Hi @haakon-e @costachris 

These look great! Merge away!

PR #370, opened 2024-03-07
Add `save_parameter_samples`

nefrathenrici:
This adds a function `save_parameter_samples` which takes samples from the given Samples distribution and saves them to individual TOML files for use in model runs.

### DIFF
--- a/src/TOMLInterface.jl
+++ b/src/TOMLInterface.jl
@@ -13,7 +13,7 @@ export save_parameter_ensemble
 export get_admissible_parameters
 export get_regularization
 export write_log_file
-
+export save_parameter_samples
 
 """
     get_parameter_values(param_dict, names)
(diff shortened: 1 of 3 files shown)


src/TOMLInterface.jl
```diff
@@ -301,6 +301,46 @@ function construct_2d_array(arr::Expr)
...
+    rng = Random.MersenneTwister(1234),
+) where {Samples <: ParameterDistributions.Samples}
+
+    save_parameter_ensemble(
+        sample(rng, distribution, num_samples),
+        distribution,
+        default_param_data::Dict,
+        save_path,
+        save_file;
+        pad_zeros = 3,
```
odunbar, line 339: Is there a reason we hard-code this to 3?

src/TOMLInterface.jl
```diff
@@ -301,6 +301,46 @@ function construct_2d_array(arr::Expr)
...
+Arguments:
+- `distribution` - ParameterDistribution{Samples} to sample from
+- `default_param_data` - Dict of default parameters to be combined and saved with
+the parameters in `param_array` into a toml file
+- `save_path` - Folder where the parameters will be saved
+- `save_file` - Name of the toml files to be generated
+- rng - The random number generator used in sampling
+"""
+function save_parameter_samples(
+    distribution::ParameterDistribution{Samples},
```
odunbar, line 325: This function would work with any distribution right? It doesn't need to be a `Samples`-type?
Unless there is a reason to restrict to this, we could allow others

odunbar (approved):
Hi @nefrathenrici 
Just a couple of small points to address then merge away

LGTM!

src/TOMLInterface.jl
nefrathenrici: As long as `sample` is defined for the given distribution, it would work. I will remove the type constraint!

src/TOMLInterface.jl
nefrathenrici: No, I set it to be an optional keyword argument.

PR #380, opened 2024-05-23
Update groups, i.e. domain and observation localization

### LINKED ISSUE #288: Sparse inversion has no verbose information
We have a verbose flag that outputs diagnostics during solves for all tools, except for Sparse inversion, these should be added for this tool too. It may be easier to centralise the call too, so that it is called within update_ensemble! in EKP and not the individual dispatched methods - just a thought.

### LINKED ISSUE #377: O3.7.5 Implement Batching for input and output groups (i.e. Domain localization and observation localization)
# Purpose
To reduce computational complextity via blocking the dependence of parameters and data. The blocking would break and update of all parameters into several small updates of partitions of the parameter space. All this would take place within each single iteration of EKP.

## Latest update: 
see PR #380 

## Example
Prototype a user interface to pass 
- parameter partitioning (via e.g. names of the parameter), NB we may wish to enforce that no two parameters can be in different groups 
- data selection (not necessarily a partition, as data can be reused to update different parameter groups parameters)
- maybe using different EK `process`/ `localizer`/ `inflation` etc. for these different update stages (all within the same iteration)

Simultaneously we should be thinking about how this interface could be used within a setting of mini-batching, where different partial observations are used in _different_ EKP iterations

## Discussed implementation

- Create an array of structs defining a conditionally dependent parameters and , passed into EKP constructor `update_groups =` in addition to current arguments
```julia
struct UpdateGroup {VV <: AbstractVector}
    u_subset::VV 
    y_subset::VV
    # process::Process # in future
    # localizer::Localizer # in future
    # inflation::Inflation # in future
end
```
- check all u_subsets form a partition of `1:p`
- if not provided create one `UpdateGroup` with u_subsets and y_subsets being the full indices
- In the `EnsembleKalmanProcess.jl` `update_ensemble!(...)` function, Replace
```julia
u = update_ensemble!(ekp,...) 
```
with 
```julia
u = zeros(param_size,ens_size)
for group in update_groups # parallelizable!
    idx =group.u_subset
    u[idx, :] = update_ensemble!(ekp, group, ...) 
end
```
### Possible hurdles
- move `push!(g,...)` out of inner `update_ensemble!` call to be outside
- where processes use priors, e.g. UKI/EKS (e.g. need to subset process.cov)
- UKI generating initial ensemble
- do we calculate the parameter dimension from a fixed quantity anywhere? or is it all deduced from passed in values

### Example problem
- Non-timescale separated coupled system with data

odunbar:
## Purpose 

Closes #377
Indirectly, Closes #288 

To review the PR and see what is going on I would suggest, 
- playing with the `example/UpdateGroups/calibrate.jl`.
- Looking at the `src/UpdateGroups.jl` file
- Looking at `src/EnsembleKalmanProcess.jl` function `update_ensemble!(...)`
- look at the docs page ["Update Groups"](https://clima.github.io/EnsembleKalmanProcesses.jl/previews/PR380/update_groups/)
 
## Content

- `UpdateGroup` object
- unit tests for `UpdateGroup`
- implemented an interface with all EKP methods, in  particular this necessitated  lifting saving of "g", and verbose flag statements out of the individual updates thus reducing a lot of duplicated code
- implemented functionality for EKI
- Constructed an example based on learning "fast" and "slow" parameters of a coupled L96 model, hopefully improved by the grouping process. 
- Updated to work with new Observations and ObservationSeries framework: In particular, a nice constructor that takes in `prior, observation, dict_in_to_out` and produces a list of `UpdateGroups` that can be passed to EKP. 
- It also now works with minibatching, by repeating index blocks for each observation in the minibatch
- Small refactor to Localizers, (instantiated without dimensionality, then dimensionality is inferred from current ensembles)
- Functionality for: ETKI,UKI,EKS,SEKI
- Docs page "Update Groups"

## Example: Multiscale lorenz 96 with j=k

Parameters (Slow)
`F`: 3-vector of amplitude, frequency and mean of a periodic forcing on slow variables
`G`: Coupling term in slow equation to the fast variables
Parameters (Fast)
`h`: Coupling term in the fast variables
`c`: Timescale separation parameter of the fast variables
`b`: nonlinearity scaling in the fast variables
With labelled observations observed "Independently" I.e. we _assume_ independent with regards to observation (probably untrue)
`<X>, <Y>, <X^2>, <Y^2>, <XY>`
one can define two update groups as
```julia
group_identifiers = Dict(
    ["F"] => ["<X>", "<X^2>"], 
    ["G", "h", "c", "b"] => ["<Y>", "<Y^2>", "<XY>"],
)
update_groups = create_update_groups(priors, observation, group_identifiers) # handles all the indexing 
```
By allowing more correlations in the observations e.g. create `<X><X^2>` and `<Y><Y^2><XY>`one can also define groups by these identifiers
```julia
group_identifiers = Dict(
    ["F"] => ["<X><X^2>"], 
    ["G", "h", "c", "b"] => ["<Y><Y^2><XY>"],
)
update_groups = create_update_groups(priors, observation, group_identifiers)
```
and pass to the EKP object with `update_groups = update_groups` flag
Spin up over a year.
 
samples of the trajectories used for data

forward map used for data (noise covaraince and true data `<X><Y><X^2><Y^2><XY>` nomalized by the noise)

Shown below is the initial and final 50-ensemble (normalized by the noise) pushed through the forward map, over the true sample +/- 2sd (grey), after 5 iterations.
- True parameters `[8.0, 2.0, 0.05555555555555555], 10.0, 10.0, 10.0, 10.0 ]`
- Left: Full calibration, `{F,G,h,c,b}` with all data `{<X><Y><X^2><Y^2><XY>}`. 
```julia
[7.864338120823016, 1.1500565821985165, 0.05883690301359528, 2.7165402443001034, 2.8197381618086217, 6.8072979930231945, 7.978078567957204]
```
- Right: calibration with grouped `{F}`, `{G,h,c,b}` and fast/slow data batches `{<X><X^2>}`, `{<Y><Y^2><XY>}`. 
```julia 
[5.093003061750248, 2.1573021824213985, 0.0060389967710004905, 1.0526243678897405, 0.8561707192010413, 8.331178239855038, 3.594166654372452]
```
 

----

### DIFF
--- a/.github/workflows/Tests.yml
+++ b/.github/workflows/Tests.yml
@@ -9,7 +9,8 @@ jobs:
     name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
     runs-on: ${{ matrix.os }}
     strategy:
-       matrix:
+      fail-fast: false #don't cancel all jobs if one fails
+      matrix:
         version:
           - 'lts' # Long-Term Support release
           - '1' # Latest 1.x release of julia
--- a/.github/workflows/Tests.yml
+++ b/.github/workflows/Tests.yml
@@ -9,7 +9,6 @@ jobs:
     name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
     runs-on: ${{ matrix.os }}
     strategy:
-        fail-fast: false
        matrix:
         version:
           - 'lts' # Long-Term Support release
--- a/.github/workflows/Tests.yml
+++ b/.github/workflows/Tests.yml
@@ -9,6 +9,7 @@ jobs:
     name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
     runs-on: ${{ matrix.os }}
     strategy:
+        fail-fast: false
        matrix:
         version:
           - 'lts' # Long-Term Support release
@@ -17,7 +18,6 @@ jobs:
           - ubuntu-latest
           - windows-latest
           - macOS-latest
-        fail-fast: false
     steps:
       - uses: styfle/cancel-workflow-action@0.9.1
         with:
--- a/.github/workflows/Tests.yml
+++ b/.github/workflows/Tests.yml
@@ -17,6 +17,7 @@ jobs:
           - ubuntu-latest
           - windows-latest
           - macOS-latest
+        fail-fast: false
     steps:
       - uses: styfle/cancel-workflow-action@0.9.1
         with:
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -82,6 +82,7 @@ pages = [
     "Learning rate schedulers" => "learning_rate_scheduler.md",
     "Prior distributions" => "parameter_distributions.md",
     "Observations and Minibatching" => "observations.md",
+    "Update Groups" => "update_groups.md",
     "Localization and SEC" => "localization.md",
     "Inflation" => "inflation.md",
     "Parallelism and HPC" => "parallel_hpc.md",
--- a/docs/src/update_groups.md
+++ b/docs/src/update_groups.md
@@ -44,7 +44,11 @@ observation_vec = []
 for i in 1:length(data_block_names)
     push!(
         observation_vec,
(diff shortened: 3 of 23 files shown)


odunbar:
Waited too long for review - but all tests, codecov and docs are passing / looking good. therefore I will merge and bear the brunt of responsibility on this one.

PR #384, opened 2024-06-17
Complete redesign of "Observations" object enabling introduction of minibatching

### LINKED ISSUE #382: Add convenient method for minibatching data
Allow users to provide a batching mechanism and list of observations and corresponding noise

## Solution (Updated)

Overhaul the existing `Observation` object and construct a new `ObservationSeries` object. 
- The `Observation`s can be combined and stacked to build new `Observation`s (akin to the `ParameterDistribution` objects). 
- Then the `ObservationSeries` will take a vector of such `Observation`s and a `Minibatcher` mechanism that will sample epochs of the provided series.
- EKP now builds and stores Observation quantities (like `obs_mean` and `obs_noise_cov` inside an `ObservationSeries` object), so all calls to `.obs_mean` and `.obs_noise_cov` will now be getter function from the `ObservationSeries` object and will provide the current minibatch.

The key objects and methods for this solution:
- The individual (possibly stacked) Observation
```julia 
struct Observation
    samples::Vector{Vector} # list of observation samples
    covs::Vector{Matrix} # list of covs for each observations
    names::Vector{String} # names of the observation samples
    indices::Vector{UnitRange} # indices of each for stacking the Observation (necessary?)
end
```
- The series of Observations for batching
```julia
struct ObservationSeries
    observations::Vector{Observation} # vector of observations
    batches::Vector{Vector} # batches grouped by epoch
    current_batch_index::Dict # holds 2 indices (epoch#, batch#) of the latest batch
    batcher<:MiniBatcher # Object to construct the next epoch of batches
    names::Vector String # holds id's for the observation samples
end
```
- Methods such as 
```julia
get_obs(os::ObservationSeries)
get_obs_noise_cov(os::ObservationSeries)
update_minibatch!(os::ObservationSeries)
```
### Some forms of the `Minibatcher`
The Batches are created by the following object, and its method:
```julia
struct MiniBatcher end
function generate_batches_for_epoch(b::MiniBatcher)
```
Simple instance for default/fixed batches by user
```julia
struct DeterministicMiniBatcher <: MiniBatcher
    batches::Vector{Vector}
end
generate_batches_for_epoch(b::DeterministicMiniBatcher) = get_batches(b)
```
Simple instance for batching into `batch_size` batches
```julia
struct  RandomMiniBatcher <: MiniBatcher
    batch_size::Int
    n_samples_per_epoch::Int
    rng::GlobalRNG
end

function generate_batches_for_epoch(b::RandomMiniBatcher) 
    N = get_n_samples_per_epoch(b)
    rng = get_rng(b)
    indices = shuffle(rng, collect(1:N))
    bs = get_batch_size(b)
    n_batches = Int(floor(N/bs))
    return  [ i < n_batches ? 
             indices[ (i-1) *batch_size+1 : i*batch_size] : # bs sized batches
             indices[ n_batches*batch_size+1 : end] # final large batch < 2*bs sized
             for i = 1:n_batches 
            ]
end
```
- The Observation object will update the batch at the end of the EKP update with a manual call to "update batch index" that will either iterate `current_batch_index`, or will make a call to the `MiniBatchers` to start a new epoch.

### LINKED ISSUE #383: Add ability to mutate key quantities such as the observation covariance matrix

### LINKED ISSUE #385: ETKI ignores timestepper
## Purpose
ETKI (and EKI) adapt to the timestep by dividing the noise covariance by the timestep
https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/fb45f75fa510691ded6f094a41bb85b4f408253d/src/EnsembleTransformKalmanInversion.jl#L113

unfortunately, unlike in EKI, ETKI does not use this quantity, instead it uses an internally stored inverse covariance. 

## Solution

Simply multiply the stored inverse covariance in `etki_update` by the timestep to pass this effect on.

odunbar:
## Purpose 

Closes #382
Closes #383
Closes #385 
Improves some of the getters in #386

Summarized by the docs [here](https://clima.github.io/EnsembleKalmanProcesses.jl/previews/PR384/observations/)
and the API docs [here](https://clima.github.io/EnsembleKalmanProcesses.jl/previews/PR384/API/Observations/)

## Content

- Remove the unnecessary Module wrapping `Observations` object. And Replaces the old and largely useless `Observation` object
- Create new `Observation` `ObservationSeries` and `Minibatcher` object. that can store `y`'s `Gamma`'s and minibatching framework to batch up epochs over them.
- EKP now always internally stores an `ObservationSeries` rather than `y` and `\Gamma` separately. 
- Observations are now accessed using the get functions that pull the stacked-`y` and blocked-`\Gamma` from the `ObservationSeries` object for the current minibatch
- `update_minibatch!(` is called at the end of each EKP step, updating the batch. At the end of the epoch, this also calls `create_new_epoch!(` for the minibatcher to create a new epoch of minibatches
- compatible with EKI, ETKI, EKS and UKI - as `Observation` stores the inverse obs noise cov too.
- learning rate schedulers compatible with minibatching
- Back-compatible with old interface passing in `y` and `Gamma` to EKP
- changes examples that used the old `Observation` object
- unit tests for all new structs
- added a docs page with small examples
- API & docstrings
- resolves the ETKI bug with the timestepper, and ensured scaling preserved in new update

----


src/EnsembleKalmanProcess.jl
```diff
@@ -163,14 +160,16 @@ function EnsembleKalmanProcess(
...
+    OS <: ObservationSeries,
 }
 
     #initial parameters stored as columns
     init_params = DataContainer(params, data_are_columns = true)
 
     # dimensionality
     N_par, N_ens = size(init_params) #stored with data as columns
-    N_obs = length(obs_mean)
+    obs = get_obs(observation_series)
```
costachris, line 171: Could be helpful to have a quick comment here or change the variable name `N_obs` to indicated whether it's number of obs in a batch or total number of obs

odunbar: Done, and added a note to the glossary.md

I think in future we will want to remove this exact line, as currently this pins the Localizer (the only place N_obs is used here) to work with a fixed minibatch size

src/EnsembleTransformKalmanInversion.jl
```diff
@@ -44,29 +48,80 @@ end
...
     g::AbstractMatrix{FT},
-    y::AbstractVector{FT},
-    obs_noise_cov::Union{AbstractMatrix{CT}, UniformScaling{CT}},
-) where {FT <: Real, IT, CT <: Real}
-    m = size(u, 2)
-    Γ_inv = ekp.process.Γ_inv
+) where {FT <: Real, IT}
 
+    y = get_obs(ekp)
+    inv_noise_scaling = ekp.Δt[end]
```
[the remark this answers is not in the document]
odunbar: This is my current dream, and we are working towards it. 

I think `get_\Delta t_final` would fit with `get_u_final` etc, I'll create an issue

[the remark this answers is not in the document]
odunbar: https://github.com/CliMA/EnsembleKalmanProcesses.jl/issues/386

docs/src/observations.md
```diff
@@ -1,23 +1,153 @@
...
+- `FixedMinibatcher(given_batches, "random")`, minibatches are fixed, but are run through in a randomly chosen order in each epoch
+- `no_minibatcher(size)`, creates a `FixedMinibatcher` with just one batch which is the epoch (effectively no minibatching)
+- `RandomFixedSizeMinibatcher(minibatch_size, "trim")`, (default `method = "trim"`) creates minibatches of size `minibatch_size` by randomly sampling the epoch, if the minibatch size does not divide into the number of samples it will ignore the remainder (and thus preserving a constant batch size)
+- `RandomFixedSizeMinibatcher(minibatch_size, "extend")`, creates minibatches of size `minibatch_size` by randomly sampling the epoch, if the minibatch size does not divide into the number of samples it will include the remainder in the final batch (and thus will cover the entirety of the data, with a larger final batch)
+
+## Observation Series
+
+One can additionally provide a vector of `names` to name each `Observation` in the `ObservationSeries` by giving using the `Dict` entry `"names" => names`
```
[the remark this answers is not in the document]
odunbar: Done

docs/src/observations.md
```diff
@@ -1,23 +1,174 @@
...
+The compiled information is given in the object:
+```julia
+yz_observation_series::ObservationSeries
+```
+As this contains many time windows, setting the names of the `ObservationSeries` objects to index the time window is a sensible identifier, for example,
+```julia
+get_names(yz_observation_series)
+> ["window_1", "window_2", ..., "window_100"]
+```
+The individual `Observation`s should refer only to the state being measured, sosuitable identifiers might be, for example,
```
eviatarbach, line 168: "sosuitable" missing space

src/EnsembleKalmanProcess.jl
```diff
@@ -429,36 +449,85 @@ end
...
-Return accelerator type of EnsembleKalmanProcess.
+Return `accelerator` field of EnsembleKalmanProcess.
 """
 function get_accelerator(ekp::EnsembleKalmanProcess)
     return ekp.accelerator
 end
 
+"""
+    get_observation_series(ekp::EnsembleKalmanProcess)
+Return `obs_noise_cov` field of EnsembleKalmanProcess.
```
eviatarbach, line 484: Wrong description

src/EnsembleKalmanProcess.jl
```diff
@@ -486,8 +555,18 @@ The error is stored within the `EnsembleKalmanProcess`.
...
     mean_g = dropdims(mean(get_g_final(ekp), dims = 2), dims = 2)
-    diff = ekp.obs_mean - mean_g
-    X = ekp.obs_noise_cov \ diff # diff: column vector
+    diff = get_obs(ekp) - mean_g
+    #   X = get_obs_noise_cov(ekp) \ diff # diff: column vector
+
+    Γ_inv = get_obs_noise_cov_inv(ekp, build = false)
+    γ_sizes = [size(γ_inv, 1) for γ_inv in Γ_inv]
+    X = zeros(sum(γ_sizes), size(diff, 2)) # stores Y' * Γ_inv
+    shift = [0]
```
eviatarbach, line 564: Why does this need to be an array?

src/EnsembleTransformKalmanInversion.jl
```diff
@@ -44,29 +48,96 @@ end
...
+
+    y = get_obs(ekp)
+    inv_noise_scaling = ekp.Δt[end]
 
+    m = size(u, 2)
     X = FT.((u .- mean(u, dims = 2)) / sqrt(m - 1))
     Y = FT.((g .- mean(g, dims = 2)) / sqrt(m - 1))
-    Ω = inv(I + Y' * Γ_inv * Y)
-    w = FT.(Ω * Y' * Γ_inv * (y .- mean(g, dims = 2)))
+    tmp = get_buffer(get_process(ekp)) # the buffer stores Y' * Γ_inv of [size(Y,2),size(Y,1)]
```
eviatarbach, line 68: This whole section of code is quite difficult to read and verbose. Any way it can be simplified?

eviatarbach (changes requested):
This is excellent, thank you for all the work on this!

I made a few comments in the code. Besides these comments, I was wondering why you removed the multiple samples of y in `Cloudy_example_eki.jl` and `aerosol_activation.jl`? It seems like it would be useful to have an example with multiple samples, especially now that they can be handled better.

src/EnsembleTransformKalmanInversion.jl
costachris: \+ 1 . If it can't be simplified, some additional comments would also be helpful (and commented out code snippets removed)

src/LearningRateSchedulers.jl
```diff
@@ -198,15 +198,9 @@ function calculate_timestep!(
...
-    if isa(ekp.obs_noise_cov, UniformScaling)
-        Γ = ekp.obs_noise_cov.λ * I(M) # converts into MxM matrix, 
-    else
-        Γ = ekp.obs_noise_cov
-    end
-    # (G(u) - E[G(u)])ᵀΓ⁻¹(G(u) - E[y])
-    D = (1 / J) * ((g .- g_mean)' * (Γ \ (g .- y_mean)))
-
+    y_mean = get_obs(ekp)
+    Γinv = get_obs_noise_cov_inv(ekp)
```
costachris, line 202: perhaps `Γ_inv` to be consistent with rest of code

src/EnsembleTransformKalmanInversion.jl
odunbar: I think in the end it's just doing the block-wise linear algebra, and using buffers to minimize the recompute.
I think I can't "simplify" the actions but I can use some functions to make it look cleaner! I'll take a look

docs/src/observations.md
```diff
@@ -1,23 +1,174 @@
...
+get_indices(full_obs)
+```
+
+## Recommended constructor: Many stacked observations
+
+Imagine the user has 100 independent data samples  for two independent variables above, where the `k`th `y` sample is = `k*ones(5)` for each `k=1:100`.
+Rather than stacking all the data together at once (forming a full system of size `100*(8+5)` to update at each step) instead the user wishes to stream the data and do updates with random batches of 5 observations at each iteration.
+
+!!! note "Why would I choose to minibatch?"
+    The memory- and time-scaling of many EKP methods is worse-than-linear in the observation dimension, therefore there is often large computational benefit to minibatch EKP updates. Such costs must be weighed against the cost of additional forward map evaluations needed to minibatching over one full epoch.
```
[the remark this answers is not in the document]
odunbar: Minibatching updates _all_ the parameter dimensions with _partial_ data. It runs another forward map evaluation with the updated parameters, and then updates _all_ the parameter dimensions against different _partial_ data. 

this is contrary to what we were trying to do with the sub-update type batching of "domain/observation"localization, where everything is done within one step

docs/src/observations.md
```diff
@@ -1,23 +1,174 @@
...
+z_obs = Observation(
+    Dict(
+        "samples" => z,
+        "covariances" => cov_z,
+        "names" => "z_$k",
+    ),
+)
+push!(hundred_full_obs, combine_observations([y_obs,z_obs]))
+end
+T = promote_type((typeof(h) for h in hundred_full_obs)...)
```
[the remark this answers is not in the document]
odunbar: It's not actually, as we do this internally

src/EnsembleKalmanProcess.jl
odunbar: It used to be that Julia would not allow mutation in a loop, maybe this changed recently...
https://discourse.julialang.org/t/loop-scope-works-differently-for-int-and-arrays-why/28172/7

odunbar: it's a pattern that is elsewhere in this code so perhaps we can clean this up another time altogether

costachris (approved):
Thanks for addressing the comments - Looks good to me

odunbar:
@eviatarbach In response to your comment, before we were not actually using more than one sample even though we gathered many, so it didn't make sense. My replacement does not effect the functionality. However you are right - we could include the multiple samples in examples - but perhaps this can be left to a future PR?

eviatarbach:
LGTM! Thank you.

eviatarbach (approved):

PR #392, opened 2024-08-05
WIP: replace defaults with cutting edge settings

odunbar:
## Content

----


docs/src/defaults.md
```diff
@@ -0,0 +1,85 @@
+## [Default Configurations](@id defaults)
+
+Sensible defaults have been chosen for the methodology to give broadly the best solution. The configurations can be revealed by adding the keyword
```
Julians42, line 3: Could be nice to add pointers to where each method is discussed. e.g., link the `TransformInversion` section to https://clima.github.io/EnsembleKalmanProcesses.jl/previews/PR392/ensemble_kalman_inversion/#etki. Not necessary though

docs/src/learning_rate_scheduler.md
```diff
@@ -13,23 +13,25 @@ We shall compare the following configurations of implemented schedulers.
 4. Misfit controlling timestep (Terminating) `DataMisfitController()` [Iglesias & Yang 2021](https://doi.org/10.1088/1361-6420/abd29b) - purple
 5. Misfit controlling timestep (Continuing beyond Terminate condition) `DataMisfitController(on_terminate="continue")` - brown
 
-One can define the schedulers as
-```julia
-scheduler = DefaultScheduler(0.5) # fixed stepsize, default values: 1
-```
-Then when constructing an EnsembleKalmanProcess, one uses the keyword argument
+!!! info "Recommended Scheduler"
```
Julians42, line 16: I agree that this is the best scheduler to use for getting the posterior but may not be the best to start with, e.g., someone's first calibration run, because the total number of iterations is not defined. I'm thinking particularly about expensive forward models.

docs/src/learning_rate_scheduler.md
```diff
@@ -13,23 +13,25 @@ We shall compare the following configurations of implemented schedulers.
...
 ```julia
 ekpobj = EKP.EnsembleKalmanProcess(args...; scheduler = scheduler, kwargs...)
 ```
-A variety of other schedulers can be defined similarly:
+Several choices are available:
 ```julia
 scheduler = MutableScheduler(2) # modifiable stepsize
 scheduler = EKSStableScheduler(numerator=10.0, nugget = 0.01) # Stable for EKS
-scheduler = DataMisfitController(on_terminate = "continue") # non-terminating
+scheduler = DataMisfitController(terminate_at = 1000) # stop at algorithm time "T=1"
```
Julians42, line 30: Assuming you meant to write `terminate_at = 1`?

docs/src/learning_rate_scheduler.md
Julians42: One suggestion could be to use this scheduler and then set a `max_iter` (assuming there's a kwarg for this?) Then the user could indicate roughly how much resources they're willing to allocate for an initial calibration.

examples/LossMinimization/loss_minimization.jl
```diff
@@ -59,7 +62,15 @@ initial_ensemble = EKP.construct_initial_ensemble(rng, prior, N_ensemble)
...
-ensemble_kalman_process = EKP.EnsembleKalmanProcess(initial_ensemble, G_target, Γ_stabilization, Inversion())
+ensemble_kalman_process = EKP.EnsembleKalmanProcess(
+    initial_ensemble,
+    G_target,
+    Γ_stabilization,
+    Inversion(),
+    scheduler = DefaultScheduler(1),
+    accelerator = DefaultAccelerator(),
+    localization_method = EnsembleKalmanProcesses.Localizers.NoLocalization(),
+)
```
Julians42, line 73: Ollie - these changes are just to prevent new defaults from kicking in correct?

Julians42 (approved):
Looks good Ollie. Left a couple of thoughts I had on the docs based on my experience, but feel free to ignore.

docs/src/learning_rate_scheduler.md
odunbar: So for this scheduler the user still submits a max iteration, and here it really will be an upper bound on the iterations. The only difference is that this will prevent further updates if the number of required iterations go beyond the end time. The benefit is the calibrations are much more robust, and have more interpretation than the fixed timestep one, and due to the adaptivity are typically more efficient.

However I'll think about whether there is a nicer message/hints that we can do to aid first-time users.

docs/src/learning_rate_scheduler.md
odunbar: Done

docs/src/defaults.md
odunbar: Done

docs/src/learning_rate_scheduler.md
odunbar: I've created a message to indicate that we return `true` from  `update_ensemble!(` I also flushed out the LRS docs.

examples/LossMinimization/loss_minimization.jl
odunbar: Yes, good spot, I added this line into the docs.

examples/LossMinimization/loss_minimization.jl
```diff
@@ -1,5 +1,8 @@
 # # Minimization of simple loss functions
 #
+# !!! info "How do I run this code?"
```
costachris, line 3: uncomment? Currently "The full code is found in ..." is outside of the "How do I run this code?" box

examples/Sinusoid/sinusoid_example.jl
```diff
@@ -1,5 +1,8 @@
 # # [Fitting parameters of a sinusoid](@id sinusoid-example)
-
+#
+# !!! info "How do I run this code?"
+#    The full code is found in the [`examples/`](https://github.com/CliMA/EnsembleKalmanProcesses.jl/tree/main/examples) directory of the github repository
```
costachris, line 4: ditto loss_minimization.jl comment

examples/SparseLossMinimization/loss_minimization_sparse_eki.jl
```diff
@@ -1,5 +1,8 @@
 # # Minimization of simple loss functions with sparse EKI
 #
+# !!! info "How do I run this code?"
+#    The full code is found in the [`examples/`](https://github.com/CliMA/EnsembleKalmanProcesses.jl/tree/main/examples) directory of the github repository
```
costachris, line 4: ditto loss_minimization.jl comment

src/EnsembleKalmanProcess.jl
```diff
@@ -65,6 +65,46 @@ struct FailureHandler{P <: Process, FM <: FailureHandlingMethod}
 end
 
 
+function default_options_dict(process::P) where {P <: Process}
+    if nameof(typeof(process)) == :Inversion
```
costachris, line 69: can also do `if isa(process, Inversion)` and similar for the remaining ifelse conditionals

costachris (approved):
Looks good! Final suggestion is to make sure this goes out as a breaking release.

src/EnsembleKalmanProcess.jl
odunbar: I think we had some issues with the parametric typing for this type of call. I can check but Ill take a look later.

examples/SparseLossMinimization/loss_minimization_sparse_eki.jl
odunbar: Thanks!! I'll try to resolve this..

docs/src/defaults.md
```diff
@@ -0,0 +1,87 @@
...
+```julia
+scheduler = DataMisfitController(terminate_at = 1)
+localization_method = Localizers.NoLocalization()
+failure_handler_method = SampleSuccGauss()
+accelerator = DefaultAccelerator()
+```
+
+## "Vanilla" settings: How to turn off features
+
+The following shows how one can recover a method with no additional features
```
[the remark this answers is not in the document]
odunbar: Done

docs/src/learning_rate_scheduler.md
```diff
@@ -7,31 +7,35 @@ We demonstrate the behaviour of different learning rate schedulers through solut
...
 scheduler = EKSStableScheduler(numerator=10.0, nugget = 0.01) # Stable for EKS
-scheduler = DataMisfitController(on_terminate = "continue") # non-terminating
+scheduler = DataMisfitController(terminate_at = 1000) # stop at algorithm time "T=1000"
 ```
 Please see the [learning rate schedulers API](@ref scheduler_api) for defaults and other details
 
-### Early termination
+## Early termination (with adaptive learning rate)
+
+When using an adaptive learning rate, early termination may be triggered when a scheduler-specific condition is satisfied prior to the final user prescribed `N_iter`.
```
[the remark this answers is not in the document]
odunbar: done

examples/LossMinimization/loss_minimization.jl
odunbar: It's actually an indenting error i think with the !!! info block, resolving...

src/EnsembleKalmanProcess.jl
odunbar: Seems to go through! I'm running tests to confirm!

PR #407, opened 2024-09-04
[WIP] Add the GNKI tools and unit testing

### LINKED ISSUE #404: Add GNKI to the toolkit
## Issue
New tool to complement EKI,EKS, etc. based on Gauss-Newton and Statistical Linearization, from [here](https://doi.org/10.48550/arXiv.2010.13299)
We will use the terminology GNKI, rather than the IEKF-SL terminology from the paper

## Solution
```[tasklist]
- [ ] https://github.com/CliMA/EnsembleKalmanProcesses.jl/issues/405
- [ ] https://github.com/CliMA/EnsembleKalmanProcesses.jl/issues/406
- [ ] Add IEKF as a comparison into at least one example
- [ ] Add a docs page describing setup, and API. Add derivation as with other EKP tools
```

### LINKED ISSUE #405: Implement the IEKF tool,

odunbar:
## Purpose 

Closes #404 
Closes #405 
Coauthored by @rgjini

## Content

- Added basic GNKI struct taking in prior
- Added update function, 
- Added some defaults for timestepper etc.
- Coverage of unit testing passes 
- Docstrings
## MISC
- Also resolves some bugs in testing where we need to  `deepcopy(Unscented(..))` processes when reusing them as they store a history internally
- Also resolves an "always-false" check statement where `nameof(typeof(X)) == name` should be `Symbol(name)`. Replaced all cases with `isa(X,name)`
- Also resolves bugs in `github-actions` where old packages are loaded instead of new ones, and resolves a related docs issue

## Example output: Sinusoid `N_ens=10`, `N_iter=5`
 Ran defaults with `timestep = 1`
i.e. 
```julia
ensemble_kalman_process = EKP.EnsembleKalmanProcess(
    initial_ensemble,
    y,
    Γ,
    GaussNewtonInversion(prior),
    localization_method = Localizers.NoLocalization(),
    accelerator = DefaultAccelerator(),
) 
```
output
and with localization + accelerator, `timestep =1` "bells and whistles"
```julia
ensemble_kalman_process = EKP.EnsembleKalmanProcess(
    initial_ensemble,
    y,
    Γ,
    GaussNewtonInversion(prior),
)
```
output

----


src/GaussNewtonKalmanInversion.jl
rgjini, line 1: The algorithm looks good to me!

rgjini (approved):
Everything looks good to me!  A couple questions: there are a few places with warnings saying certain lines are not covered by the tests, is that normal?  I was also curious what ekp.verbose means?  Is this something all the algorithms have?

odunbar:
> Everything looks good to me! A couple questions: there are a few places with warnings saying certain lines are not covered by the tests, is that normal? 

This means that we added some new lines of code that are untested - we require ~90% of the new code to be covered as a soft-rule so we can add a couple more tests to cover anything important that was missed

> I was also curious what ekp.verbose means? Is this something all the algorithms have?

Yes, if we add `verbose = true` when creating the EKP object it prints some basic diagnostic-type information about the convergence

By the looks of it, I shall 
- [ ] improve the test coverage
- [ ] look at why the windows tests fail. It's nothing we did, rather it is a some package issue in the pipeline.

odunbar:
All resolved - merging on tests passing and docs building

PR #417, opened 2024-10-16
Add new spatially dependent lorenz-forcing calibration

odunbar:
## Content

----

### DIFF
--- a/examples/Lorenz/Lorenz_example_spatial_dep_forcing.jl
+++ b/examples/Lorenz/Lorenz_example_spatial_dep_forcing.jl
@@ -1,0 +1,163 @@
+include("GModel.jl") # Contains Lorenz 96 source code
+
+# Import modules
+using Distributions  # probability distributions and associated functions
+using LinearAlgebra
+using StatsPlots
+using Plots
+using Random
+using JLD2
+
+# CES 
+using EnsembleKalmanProcesses
+using EnsembleKalmanProcesses.DataContainers
+using EnsembleKalmanProcesses.ParameterDistributions
+
+const EKP = EnsembleKalmanProcesses
+
+#######
+#Running the Lorenz model:
+# Solve the Lorenz 96 system 
+# Inputs: F: scalar forcing Float64(1), Fp: initial perturbation Float64(N)
+# N: number of longitude steps Int64(1), dt: time step Float64(1)
+# tend: End of simulation Float64(1), nstep:
+
+G(θ) = H(Ψ(θ,x₀,t₀,t₁))
+y = G(θ) + η
+
+# This will change for different Lorenz simulators
+struct LorenzConfig{FT <: Real}
+    dt::FT
+    T::FT
+end
+
+# This will change for each ensemble member
+struct EnsembleMemberConfig{VV <: AbstractVector}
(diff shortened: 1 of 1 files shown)


rgjini:
Work in progress (WIP)

rgjini:
Debugging Notes: 
- x0 currently initialized as type VorM, not sure what that is, but I'm getting an error and will change it for now

rgjini:
The spin-up period in the statistics function is working properly: 
- I checked the trajectory of x(t) and it seems like the initial spinup is a little less than 2 total time steps.  Multiplied by 2, gives 4 time units for spinup
- I looked at heatmaps in python, attached is an example or a trajectory 
spinup_trajectory

rgjini:
I finished up the example and everything seems to be working! I fixed the random seed and ensemble size for loops, so those are both working.  I kept the amount of different ensemble sizes and random seeds low for now, the code does take a while to run (maybe 10 minutes).  I also have a few plots I made at the end.  Two plots are of the final optimized solution compared to the true parameters for different ensemble sizes and the last one is supposed to be how many iterations each algorithm took for each ensemble size.  For some reason, the last plot isn't looking how I would expect, but it might be related to an issue I ran into that I will describe in another comment.  Here are some of the plots I made!:
Screenshot 2025-01-06 at 3 00 15 PM
Screenshot 2025-01-06 at 2 59 03 PM
Screenshot 2025-01-06 at 2 58 26 PM

rgjini:
I ran into an issue running through the different algorithms for different ensemble sizes and random seeds.  After the first iteration, I think UKI doesn't get reinitialized and starts at the same place it ended from the last time the algorithm was run.  Every time after the first time UKI runs, the algorithm's RMSE is low for its initial ensemble and then doesn't even run through the algorithm.  I think I can try to change where the UKI method gets initialized, but I wanted to bring it up here because I'm not sure if that is the expected behavior when all the other algorithms seem to be reinitialized when running the different random seeds and ensemble sizes.

rgjini:
I finished including all the changes we talked about into the example script and I believe everything is ready to go!  Attached are some new plots with the parameter estimates, the model output from the ensemble means compared with the data (and associated errors), and a plot showing how many forward runs each of the algorithms take to converge: 
fow_run_comparison
standarddeviations90
averages90
solution90
solution50

odunbar:
Hi @rgjini These look awesome, I have rerun the simulations, and got the results below! I have also rebased to main so merging is possible again!

If you are pleased with this then I am happy to merge

PS: I think it is still worth discussing the exact observations (i.e. the form of "G") at a later date, as I think this experiment does very long averages to eliminate an observational noise that EKP methods don't need to avoid and then we add artificial noise back on to the observations.  In particular, the current setup way we have `G` and the `noise` I think only works with `T~500`,  but really we could do maybe have good learning with something as small as `T~10`.

rgjini:
@odunbar Thank you!  The plots look great to me, so I'll go ahead and merge :) 

I agree, I think it's worth discussing more.  If you are still free at 10:30 before we meet with Matti and Tapio, maybe we can discuss the observations more then? I'm also happy to cancel that meeting, and we can either discuss it during our meeting at 11 or in a couple of weeks?

PR #420, opened 2024-11-19
Use `eltype` instead of `promote_type`

glwagner:
Otherwise we can get a StackOverflow error with large `samples`, eg

```julia
ERROR: LoadError: StackOverflowError:
Stacktrace:
 [1] promote_type(::Type, ::Type, ::Type, ::Type, ::Vararg{Type}) (repeats 600 times)
   @ Base ./promotion.jl:299
 [2] Observation(obs_dict::Dict{String, Any})
   @ EnsembleKalmanProcesses ~/.julia/packages/EnsembleKalmanProcesses/hifDM/src/Observations.jl:125
 [3] include(fname::String)
   @ Base.MainInclude ./client.jl:489
```

(And this may only be an issue on mac!)

### DIFF
--- a/src/Observations.jl
+++ b/src/Observations.jl
@@ -122,7 +122,7 @@ function Observation(obs_dict::Dict)
     if !isa(samples, AbstractVector) # 1 -> [[1]]
         snew = [[samples]]
     else
-        T = promote_type((typeof(s) for s in samples)...)
+        T = eltype(samples)
         samples_tmp = [convert(T, s) for s in samples] # to re-infer eltype  (if the user makes an eltype of "Any")
         if !isa(samples_tmp, AbstractVector{<:AbstractVector}) # [1,2] -> [[1,2]] 
             snew = [samples]


glwagner:
Reproducer

```julia
julia> metadata = Dict("samples" => rand(2000),
                       "covariances" => EKP.I,
                       "names" => "surface temperature" )
Dict{String, Any} with 3 entries:
  "names"       => "surface temperature"
  "samples"     => [0.72443, 0.237406, 0.353835, 0.315662, 0.282328, 0.111403, 0.515477, 0.769607, 0.226976, 0.579689  …  0.497924, 0.55258, 0.730444, 0.775595, 0.58…
  "covariances" => UniformScaling{Bool}(true)

julia> y = Observation(metadata)
ERROR: StackOverflowError:
```

odunbar (approved):
LGTM - and this fixes the LTS test failures.

PR #436, opened 2024-12-18
Bugfix a lost Delta t scaling for ETKI

odunbar:
## Purpose 

Bugfix to bug introduced in PR #380,

## Content

----


odunbar:
Passes all tests, and I have confirmed results match with example `Accelerators/ExpSinusoid`  therefore OK'ing merge

PR #438, opened 2025-01-13
January clean-up

### LINKED ISSUE #408: A few potential README improvements
1. Move the "citing us" to the bottom of the README

Because people don't need to cite until they actually start using the code!
Also, add a link to the JOSS. The README just says "cite our article" but doesn't give a link.

2. Add a code snippet illustrating usage

This could be part of the "getting started" section. It really helps ground the information that we receive --- what are we trying to do here? --- otherwise the flow / decision charts can be disorienting if this is the very first time you've seen the package.

Hopefully it can be quite short, eg 20 lines or less.

### LINKED ISSUE #409: Doubled warnings
https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/3a32bd18ba8db217dd00699ad3e93067c1742754/src/EnsembleKalmanProcess.jl#L206

This should be a nested if statement, in the `N_par<10` statement in preceding lines otherwise the warning double-up

### LINKED ISSUE #413: Alternative approximation of std-approx in SECNice()
## Issue
in SECNice() we must approximate the standard deviation of correlations for a given sample size (ensemble size). Here done via Fisher transform & sampling
https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/08c735e8440a0ad2e5a54da1e01fc627e627391c/src/Localizers.jl#L367

This works fine, but is too slow in general to do for every correlation matrix entry, therefore we precompute this on a grid.

Matthias Morzfeld (UCSD) has sent over another, even simpler approximation, applicable elementwise and needs no precomputation
```
sig_{ij} = ( 1-\rho_{ij} )/ \sqrt{N_ens}
```

## Solution
Let's implement it!

### LINKED ISSUE #421: Multiple ensemble members identical - request for better error message
Currently if columns(G_ens) are equal then on solve a `LinearAlgebra SingularException` is thrown, this should be easy to catch, (against the size of the noise) and a better error message should be produced

### LINKED ISSUE #423: Remove the Unnecessary inflexibility in building the sampler
https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/0c742b0988c80f0fbda3ba54428114a91df047a5/src/Observations.jl#L790-L791

### LINKED ISSUE #433: Docs still mention bors
Seen here: https://clima.github.io/EnsembleKalmanProcesses.jl/dev/contributing/#The-merge-process

odunbar:
## Purpose 

Sort a variety of low-priority tasks that have been accruing
- Closes #433
- Closes #409
- Closes #421
- Closes #413
- Closes #423
- Closes #408 

## Content

- Replaces mention of bors with squash and merge request
- `elseif` in warning about `N_ens`/`N_par` to remove duplication
- added column check for G_ens in update_ensemble, if non-NaN columns are overlapping, a warning is produced indicating this prior to LinearAlgebra issues. Unit tested as well.
- improves readout for `FailureHandler` and `Localizer` when verbose is used (before this did not show the actual method used)
- new SECNice approximation is accurate to 10^-4 on average to the prior approximation and requires no precompute!
- Updated Localizers example to new defaults and for readability
- `get_obs` obtains the length of a vector by summing piece lengths, not estimating it by multiplying the first vector length
- Added copy-pasteable(!) example into the Readme complete with forward map data etc. for users to play with. 
- Moved citation down and added a link

## Validation
- Unit tests for the G_ens multiple columns the same
- Verified SECNice accuracy with localizer example and direct approximation comparison
- Readme can be viewed [here](https://github.com/CliMA/EnsembleKalmanProcesses.jl/tree/orad/cleanup-Jan25)
- All checks/tests pass

----


odunbar:
As I have condensed a few small changes in here - I will merge when all tests, coverage and other checks pass. 
I will reach out specifically for the README review after the merge has occured so we can build and improve on this initial attempt in a future PR and it does not hold up other bug-fixes.

PR #444, opened 2025-02-11
Impose prior option (a.k.a finite-time vs infinite-time EKI)

### LINKED ISSUE #437: Include Impose-prior versions of EKP. (a.k.a infinite-time EKI)
## Issue
This results from discussions 4.4 and 4.5 in https://arxiv.org/pdf/2209.11371 
The current methodology for EKI / ETKI is built on the formulation of finite-time transport, which provides a nice fast algorithm but with two restrictions 
(1) The solution interpretation is only approximately Bayesian only at `T=1`,  (Good if we don't trust priors and so iterate beyond)
(2) The initial ensemble must be sampled from the prior (Bad if prior is very broad / poorly sampled / model is unstable)

The infinite-time variant on the other hand has other properties
(1) It reaches the Bayesian solution at T = infty (good if we do trust priors and want a proper Bayesian solution)
(2) The initial ensemble and the prior are decoupled

## Current set-up
The infinite-time variant, can be set-up with modifications from the finite-time setup. (currently can be done in our package via:)
(1) constructing  an extended forward map and data such that
- `G(u) -> [G(u),u]'`
- `y -> [y,m]'`
- `Gamma -> [Gamma 0; 0 C]`

_addresses the incorrect minimum as the MLE_ 

(2) use covariance inflation with the current covariance matrix: _[Check the exact option]_
`update_ensemble1(... ,additive_inflation = true, additive_inflation_cov = cov(u,dims=2))`

_addresses collapse resulting in poor convergence_

Particularly with (1) this is quite restrictive.

## Possible Solution
- The user can optionally add a prior in the `Inversion` and/or `TransformInversion` object
- an `impose_prior=true` flag will be used to extend the state inside the EKI/ETKI update
- investigate whether inflation should be added internally, else is it better to prompt the user to add the correct covariance inflation to ensure quick convergence

## Misc
- compute_error! will need to include the prior term
- adaptive scheduler computations may require the prior term
- and possible index issues for update-groups

odunbar:
## Purpose 

Closes #437 

Note the terminology and design choices are motivated by Section 4.5 of [v3] https://arxiv.org/pdf/2209.11371

## Content

- Add prior into `Inversion(prior)` and `TransformInversion(prior)`
- Add the corresponding extension of the state to update the loss function with the prior component
- Automatically add `multiplicative_inflation` with default size `1e-3*min(1, 1/Delta t)` by default for EKI, and 0.0 for ETKI, if `get_impose_prior(process)` is true  - users may override this at the `update_ensemble!()` stage, or with the `default_multiplicative_inflation` keyword when creating the process.
- Added a new experiment `loss_minimization_finite_vs_infinite_ekp.jl` (see results below) in particular i also compute contours of the actual posterior density to overlay convergence
- Added unit tests for the getters, scalability of ETKI, and ability to converge toward posterior mean without starting at the prior

## Misc
- Add issue to create some docs notes about the implementation in the EKI and ETKI page. (issue #450)

## Results: (Mar 4, 25)
**ALL the following are plotted with a deterministic "G" on the left, stochastic "G" on the right. The infinite time ensembles are initialized outside the prior.**

finite-time inversion stopped at "T=1"
 

infinite-time inversion until "T=100"
 

finite-time transform-inversion stopped at "T=1"
 

infinite-time transform-inversion until "T=100"
 

sampler for longer times
 

and plots of error in ensemble mean to the true parameter over time (ensemble std in dashed line)
 

----


odunbar:
- Unit tests all pass ETKI computational scaling remains,
- Codecov is 100% on diff
- Example results are promising
-

rgjini:
I went through the EKI and ETKI code updates and everything looks good on my end!  I ran through your examples in the LossMinimization folder and everything works well.  I was also able to generate the gifs and images from the examples!

odunbar:
merging on success of tests/review

PR #462, opened 2025-03-11
Improve DMC scheduler scaling with large observation dimension

### LINKED ISSUE #461: DMC scales poorly with ETKI
## Purpose
https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/4be1b3fcf6404a318486b76feaf924cc5628b1c8/src/LearningRateSchedulers.jl#L283

In computing the timestep, we compute and store the inv. square root of the matrix, and correct positive definiteness. In doing so we build the matrix leading to terrible scaling for high dimensional output space

## Possible solution
Make use of `get_obs_noise_cov_inv(ekp, build=false)`, and apply the `sqrt` to the blocks. This in particular will maintain `Diagonal` types etc. and will cause memory issues for large matrices.

odunbar:
## Purpose 

- Closes #461 

## Content

- Compute the potential in DMC using `get_obs_noise_cov_inv(ekp, build=false)` rather than `get_obs_noise_cov(ekp)` and so avoiding building the full inverse
- remove the stored `sqrt_inv_\Gamma` previously required (as recently we now store the inverses in reduced form in EKP)
- Add `DataMisfitController()` as the scheduler for the ETKI scaling unit-test

----

### DIFF
--- a/src/LearningRateSchedulers.jl
+++ b/src/LearningRateSchedulers.jl
@@ -118,8 +118,7 @@ The user may also change the `T` with `terminate_at` keyword.
 
 $(TYPEDFIELDS)
 """
-struct DataMisfitController{FT, S} <:
-    LearningRateScheduler where {FT <: AbstractFloat, S <: AbstractString}
+struct DataMisfitController{FT, S} <: LearningRateScheduler where {FT <: AbstractFloat, S <: AbstractString}
     "the current iteration"
     iteration::Vector{Int}
     "the algorithm time for termination, default: 1.0"
@@ -283,7 +282,7 @@ function calculate_timestep!(
     else
         scheduler.iteration[end] += 1
     end
-    
+
     n = scheduler.iteration[end]
     sum_Δt = (n == 1) ? 0.0 : sum(get_Δt(ekp))
     sum_Δt_min1 = (n <= 2) ? 0.0 : sum(get_Δt(ekp)[1:(end - 1)])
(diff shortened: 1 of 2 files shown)


src/LearningRateSchedulers.jl
```diff
@@ -313,7 +306,20 @@ function calculate_timestep!(
 
     y_mean = get_obs(ekp)
 
-    Φ = [0.5 * norm(inv_sqrt_Γ * (g[:, j] - reshape(y_mean, :, 1)))^2 for j in 1:J]
+    # efficiently compute [0.5 * norm(inv(sqrt(Γ)) * (g_j - y))^2 for j in 1:J]
```
nefrathenrici, line 309: I don't really understand what is happening in the loop below with `shift` but I trust that this comment is accurate.

nefrathenrici (approved):
This looks good to me, thank you for adding more scaling tests.

src/LearningRateSchedulers.jl
odunbar: First, I have checked that the errors are within `~10^-13`, which actually comes due to the error in taking the uneccessary matrix square root in the old code.


Second, as a brief note to explain. 
- We can rewrite the `norm(sqrt(A)*(g_j-y))^2` as `(g_j - y)' * A * (g_j-y)`, 
- We compute `A*(g_j - y)` and dot this with `(gj-y)`,
- Rather than doing this for every `j` separately do it in one sweep with matrix algebra

The matrix product looks gross, as we also replace a matrix `A` with block matrices `[A1,A2,A3...]`. And therefore when performing a multiplication, `Ax` we have to multiply `A1*x[idx1,:]` and `A2*x[idx,:]`, thus we need to keep track of indices

PR #463, opened 2025-03-11
New Quantification of member failure, and more flexible NaN handling

### LINKED ISSUE #460: Failure-Handling refinement (a.k.a NaN handling)
## Currently
Failure handlers will, for each method, identify NaNs in the matrix `g` passed into update_ensemble.
- If a NaN is found, the entire column of `g` is considered failed. And the ensemble member is removed+replaced in the update.

For a model producing even just one NaN in many ensemble members, all data is lost from all members. We would like to broaden the flexibility to allow for easier interfacing with models of this type, so that not so much is lost.

## Proposed solution via preprocessing within `update_ensemble!`
given a matrix `g` containing NaNs, and given a user-specified accepted tolerance (e.g. 10%)
1. The code checks if any column is exceeds the tolerance of NaN values, such members are flagged as inadmissible. (i.e. flagged for removed+replaced as before in the update)
2. Of the admissible columns, any NaNs are replaced by finite values of the ensemble-mean (without NaNs) over the row.  The number of NaNs should be reported in some manner, 
3. If all members of a row are NaN the value of the observation itself is imputed (and so it has minimal contribution to the loss function) but a warning is thrown. A possible example is to use `y[i]`(+small noise?) to replace `g[i,:]`
4. The code updates as before

## Example
If you have dimension-50 output, with 20 ensemble members you set `nan_tolerance=0.1` then,
- A column of `g` with 10 NaNs  (corresponds to fraction of 0.2 ) will be considered a failed particle
- A column of `g` with 4 NaNs (corresponds to a fraction of 0.08) will be considered successful, and the 4NaN values will be filled by the ens-mean of each row (exc NaNs)
- A row `g[i,:]` with all NaNs will be replaced with say, `g[i,:] = y[i]`

odunbar:
## Purpose 

Closes #460 

## Content

- introduced nan preprocessing in line with issue #460
- added unit tests
- when ekp.verbose = true, produce details of NaNs found. false provides less information RE failed NaN rows, and # imputed values
- Allow users to define their fixed vector of values `nan_row_values` as a sensible replacement if they prefer alternate behaviour

## Example imputation from unit testing
Take a matrix `g`
```
g 
7×4 Matrix{Float64}:
 NaN           0.992373   NaN           -0.200025
  -1.37133     0.0527899  NaN           -1.15243
 NaN         NaN          NaN          NaN
 NaN         NaN          NaN          NaN
  -2.28799   NaN            0.614214     1.54217
   0.582582    0.0744582   -1.37219      0.253971
   0.620447    0.441834     0.0641471    0.490588


```
with `nan_tolerance = 0.5`, `nan_row_vals = collect(1:7)`, and then the imputation looks like
```
g_new = impute_over_nans(g, nan_tolerance, nan_row_vals, verbose=true)
┌ Warning: In forward map ensemble g, detected 12 NaNs. 
│ Given nan_tolerance = 0.5 to determine failed members: 
│ - Ensemble members failed:       1 
│ - NaNs in successful members:    8 
│ - rows index set for imputation: [1, 3, 4, 5] 
│ - rows index entirely NaN:       [3, 4]
[ Info: Imputed 8 NaNs
7×4 Matrix{Float64}:
  0.396174   0.992373   NaN          -0.200025     # `g[1, 1]`          -> `mean(g[1, [2, 4] ])`
 -1.37133    0.0527899  NaN          -1.15243
  3.0        3.0        NaN           3.0          #  `g[3, [1, 2, 4]]` -> `3.0`
  4.0        4.0        NaN           4.0          #  `g[4, [1, 2, 4]]` -> `4.0`
 -2.28799   -0.0438687    0.614214    1.54217      #  `g[5, 2]`         -> `mean(g[5, [1, 3, 4])`
  0.582582   0.0744582   -1.37219     0.253971
  0.620447   0.441834     0.0641471   0.490588

```

Within EKP:
-  `nan_tolerance` is a keyword given to `EnsembleKalmanProcess`
-  `nan_row_values` is a keyword given to `EnsembleKalmanProcess` for a user defined value (overriding `get_obs(ekp)` default
-  `verbose` is taken to be `ekp.verbose`

----

### DIFF
--- a/docs/src/API/EnsembleKalmanProcess.md
+++ b/docs/src/API/EnsembleKalmanProcess.md
@@ -39,6 +39,7 @@ construct_initial_ensemble
 update_ensemble!
 sample_empirical_gaussian
 split_indices_by_success
+impute_over_nans
 list_update_groups_over_minibatch
 ```
 
--- a/docs/src/API/EnsembleKalmanProcess.md
+++ b/docs/src/API/EnsembleKalmanProcess.md
@@ -35,6 +35,8 @@ get_Δt
 get_failure_handler
 get_localizer
 get_localizer_type
+get_nan_tolerance
+get_nan_row_values
 construct_initial_ensemble
 update_ensemble!
 sample_empirical_gaussian
--- a/src/EnsembleKalmanProcess.jl
+++ b/src/EnsembleKalmanProcess.jl
@@ -42,7 +42,7 @@ abstract type Accelerator end
 "Failure handling method that ignores forward model failures"
 struct IgnoreFailures <: FailureHandlingMethod end
 
-""""
+"""
     SampleSuccGauss <: FailureHandlingMethod
 
 Failure handling method that substitutes failed ensemble members by new samples from
@@ -964,8 +964,7 @@ Given nan_tolerance = $(nan_tolerance) to determine failed members:
 - Ensemble members failed:      $(sum((sum(nan_loc, dims=1) .> tol))) 
 - NaNs in successful members:   $(sum(nan_loc[:,not_fail])) 
 - row index set for imputation: $(rows_for_imputation) 
-- row index entirely NaN:      $(rows_all_nan)
-    """
+- row index entirely NaN:      $(rows_all_nan)    """
         else
             if length(rows_all_nan) > 0
                 @warn "Detected rows entirely NaN: $(rows_all_nan)"
--- a/src/EnsembleKalmanProcess.jl
+++ b/src/EnsembleKalmanProcess.jl
@@ -175,7 +175,7 @@ struct EnsembleKalmanProcess{
     LRS <: LearningRateScheduler,
     ACC <: Accelerator,
     VV <: AbstractVector,
-    NorVV <: Union{Nothing, AbstractVector}
+    NorVV <: Union{Nothing, AbstractVector},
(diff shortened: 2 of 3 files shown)


odunbar:
> When replacing entire row with observation vector y, does that influence the error quantification? (I imagine it is seen as a "perfect row", since it is equal to obs) Is that an issue? (does that make sense?)

Indeed this is the most challenging setting, where effectively the algorithm has nothing to work with.  In this case I think there are a couple of options, 
- Replace with previous iterations ensemble mean for the same value. This to me was a bit unsatisfying, as it gives a biased result, only works after iteration N=1, and only works if that previous row also didnt NaN
- Replace with data `y`, (as we have done here)

I accept that this will "improve the error computed" and that may lead to user confusion. I also admit that I do not know an optimal solution here. Improvements I can add to help:
- [ ] Add a warning for number of total failed rows even without  `ekp.verbose=true` (as we do with the failure handler) so users understand we are having to modify the sites heavily.
- [ ] Allow users to define their fixed vector of values as a sensible replacement if they prefer alternate behaviour

odunbar:
Approved in offline conversations. Merging on successful testing and codecov...

AlexisRenchon (approved):

PR #469, opened 2025-03-14
New UTKI variant `TransformUnscented`

### LINKED ISSUE #466: introduce UTKI, high-dimension scaling UKI

### LINKED ISSUE #467: UKI/UTKI currently do not have update-groups implemented

### LINKED ISSUE #468: UKI in 1 dimension breaks

odunbar:
## Purpose 

Closes #466 
Closes #467
Closes #468

## Content

- Added `TransformUnscented` inspired from our EKTI based on using the Woodbury identity to perform the update linear algebra in the ensemble space not the data space in `src/UnscentedTransformKalmanInversion.jl`(i.e. from EKTF  (https://journals.ametsoc.org/view/journals/mwre/129/3/1520-0493_2001_129_0420_aswtet_2.0.co_2.xml) - NB largely it is a similar formulation, with the main difference being the need to build "weighted" perturbations from the ensemble, not just 1/sqrt(N-1) * difference. 
- Share many of the current `Unscented` functions, with `TransformUnscented` 
- UKI & UTKI implementation with update-group indexing
- fixed `Unscented(prior)` initialization for 1-dim parameter prior
- full suite of unit tests (verifying utki and uki produce similar values)
- Added UTKI into the loss_minimization example below.

### Misc
- added "kwargs..." into update_ensemble for all methods
- fixed small index bug for etki indexing (also in utki) with update-groups
- Removed a warning for too-small ensemble size that is inappropriate for unscented methods
- Add utilities to safely multiply `lmul` and `lmul!` by the observation noise, it's inverse. 

## Examples: Examples, check with UKI
- on `examples/LossMinimization/loss_minimization_finite_vs_infinite_ekp.jl` using `impose_prior=true`  
Order: Left (smooth loss), UKI & UTKI, Right (noisy loss) UKI & UTKI
   
Both unscented variants (UKI/UTKI) behave the same on smooth loss and noisy loss (NB `impose_prior=true` has slight bias towards prior compared with e.g. EKS or other prior-imposing (a.k.a `X-infinite` variants)
 

-  on `examples/Accelerators/ExpSinAccelerated/` **Left = UKI, Right UTKI**
 

## Examples: Scaling 
- from unit tests solving problems with increased dimension and adding in regular UKI for comparison:
```
┌ Info: 6 iterations of experiment uki with 1000-dim observation took 0.7976522949999998 seconds. 
└  (avg update: 0.13294204916666663)
┌ Info: 6 iterations of experiment utki with 1000-dim observation took 0.005414058999999999 seconds. 
└  (avg update: 0.0009023431666666665)

Info: 4 iterations of experiment uki with 10000-dim observation took 109.81469675299999 seconds. 
└  (avg update: 27.453674188249998)
┌ Info: 4 iterations of experiment utki with 10000-dim observation took 0.099600575 seconds. 
└  (avg update: 0.02490014375)

[UKI not attempted]
┌ Info: 3 iterations of experiment utki with 100000-dim observation took 0.48207968700000003 seconds. 
└  (avg update: 0.16069322900000002)

[UKI not attempted]
┌ Info: 2 iterations of experiment utki with 1000000-dim observation took 5.737153102 seconds. 
└  (avg update: 2.868576551)

```

----


src/UnscentedTransformKalmanInversion.jl
```diff
@@ -0,0 +1,828 @@
...
+     automatically.
+  - `prior_mean`: Prior mean used for regularization.
+  - `prior_cov`: Prior cov used for regularization.
+  - `sigma_points`: String of sigma point type, it can be `symmetric` with `2N_par+1` 
+     ensemble members or `simplex` with `N_par+2` ensemble members.
+  
+$(METHODLIST)
+"""
+mutable struct TransformUnscented{FT <: AbstractFloat, IT <: Int} <: Process
+    "an interable of arrays of size `N_parameters` containing the mean of the parameters (in each `uki` iteration a new array of mean is added), note - this is not the same as the ensemble mean of the sigma ensemble as it is taken prior to prediction"
```
tapios, line 57:
```suggestion
    "an iterable of arrays of size `N_parameters` containing the mean of the parameters (in each `uki` iteration a new array of mean is added), note - this is not the same as the ensemble mean of the sigma ensemble as it is taken prior to prediction"
```

src/UnscentedTransformKalmanInversion.jl
```diff
@@ -0,0 +1,828 @@
...
+    # compute cholesky factor L of x_cov
+    local chol_xx_cov
+    try
+        chol_xx_cov = cholesky(Hermitian(x_cov)).L
+    catch
+        _, S, Ut = svd(x_cov)
+        # find the first singular value that is smaller than 1e-8
+        ind_0 = searchsortedfirst(S, 1e-8, rev = true)
+        S[ind_0:end] .= S[ind_0 - 1]
+        chol_xx_cov = (qr(sqrt.(S) .* Ut).R)'
```
tapios, line 409: Do you need the QR factorization here? You only really need this to be the sqrt of the covariance matrix, right? So it doesn't need to be triangular, does it?

odunbar: As this is not something we have changed for the UTKI (it was copied from the implementation in UKI) . I will create a separate issue for it! I do agree though.

src/UnscentedTransformKalmanInversion.jl
```diff
@@ -0,0 +1,474 @@
...
+    ) where {FT <: AbstractFloat, IT <: Int}
+
+Construct an TransformUnscented Inversion Process.
+
+Inputs:
+
+  - `u0_mean`: Mean at initialization.
+  - `uu0_cov`: Covariance at initialization.
+  - `α_reg`: Hyperparameter controlling regularization toward the prior mean (0 < `α_reg` ≤ 1),
+  default should be 1, without regulariazion.
```
eviatarbach, line 34: Typo "regulariazion". Also, maybe change to "default is 1"?

src/UnscentedTransformKalmanInversion.jl
```diff
@@ -0,0 +1,474 @@
...
+        throw(ArgumentError("sigma_points type is not recognized. Select the option \"symmetric\", \"simplex\"... "))
+    end
+
+    N_par = size(u0_mean, 1)
+    # ensemble size
+
+    mean_weights = zeros(FT, N_ens)
+    cov_weights = zeros(FT, N_ens)
+
+    if sigma_points == "symmetric"
```
eviatarbach, line 138: All of this logic is shared with regular UKI, is it not? It would be good to not duplicate the code if possible.

src/UnscentedTransformKalmanInversion.jl
```diff
@@ -0,0 +1,474 @@
...
+    end
+    return FailureHandler{TransformUnscented, SampleSuccGauss}(failsafe_update)
+end
+
+"""
+$(TYPEDSIGNATURES)
+
+UKI analysis step  : g is the predicted observations  `Ny x N_ens` matrix
+"""
+function update_ensemble_analysis!(
```
eviatarbach, line 354: Maybe I'm misunderstanding, but is it necessary to have both `update_ensemble_analysis!` and `succ_gauss_analysis!`? Seems to be a lot of code duplication.

eviatarbach (commented):
Looks great! I just have some comments about the code structure/duplication.

src/Observations.jl
```diff
@@ -1051,6 +1051,75 @@ function lmul_without_build!(out, A, X::AVorM, idx_triple::AV) where {AVorM <: A
     end
 end
 
+function lmul_sqrt_without_build(A, X::AVorM) where {AVorM <: AbstractVecOrMat}
```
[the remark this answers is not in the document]
odunbar: It seems like if you have an `SVD` object, for example, then `LinearAlgebra` can offer `inv()` and I think `\`. But it doesn't offer left multiplication, or things like `size()`. We also can't extend the operations due to type piracy, so I had to create several functions.

src/EnsembleKalmanProcess.jl
```diff
@@ -762,6 +775,14 @@ function lmul_obs_noise_cov!(
     return lmul_obs_noise_cov!(out, get_observation_series(ekp), X, idx_set)
 end
 
+function lmul_sqrt_obs_noise_cov!(
```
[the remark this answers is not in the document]
odunbar: I am going to remove the sqrt_obs_noise_cov ones I think, as although they could be useful, we don't actually use them. but we do use the `lmul_obs_noise_cov_inv` ones. 

I like to use them as we can use the same logic for ETKI and UTKI. And also if they are their own function i can write nicer unit tests to ensure we really get this piece correct.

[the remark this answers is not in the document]
odunbar: [Removed]

src/UnscentedTransformKalmanInversion.jl
odunbar: Done

src/UnscentedTransformKalmanInversion.jl
odunbar: Mostly yes, I have changed it so we now build the `Unscented` object with it's constructor (avoiding duplicates) then we move the fields over to `TransformUnscented`

src/UnscentedTransformKalmanInversion.jl
odunbar: I think this is something for a future PR. I have been unhappy with the way we have implemented the failure handler. but basically this is the reason for this. It is much broader than this method though, so I think it will have to persist for now.

src/UnscentedTransformKalmanInversion.jl
odunbar: Done

src/UnscentedTransformKalmanInversion.jl
odunbar: Created an issue.

eviatarbach:
Thank you! LGTM!

PR #470, opened 2025-03-15
ETKI (infinite-version) bugfixes and clarity of logic

odunbar:
## Purpose 

- I noticed that the example `loss_minimization_finite_vs_infinite.jl` had changed behaviour from v2.2.0 to v2.3.0. This bug-fixes the indexing issue and clears-up and hides the logic to prevent future index errors.

**v2.3.0** results not comparable with PR #444 (see yellow dashed line - showing ETKI infinite ensemble not collapsing)
 

## Content

- index-shift error fixed, 
- replaced `onci_idx` logic 
- Bugfix prior not being pulled in lmul function. (manually apply prior after)

## Report: 
**New** results are now matching those of PR #444
 

----

### DIFF
--- a/src/EnsembleKalmanProcess.jl
+++ b/src/EnsembleKalmanProcess.jl
@@ -757,9 +757,9 @@ function lmul_obs_noise_cov!(
     out::AM,
     ekp::EnsembleKalmanProcess,
     X::AVorM,
-    idx_triple::AV,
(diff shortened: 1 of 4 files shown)


odunbar:
Bugfixes only for ETKI, tests pass and example resolved

PR #479, opened 2025-04-10
New compat for Optim >=1.12

odunbar:
## Purpose 

- Fixes failing tests

## Content

- x_tol is deprecated, so must now use `reltol`/`abstol`. This will break compatability for older versions of Optim I expect without an unpleasant catch

----


odunbar:
We will pass on test success. It is easy to backtrack if necessary in future, resolving failing tests in the mean-time with a try-catch.

PR #486, opened 2025-04-21
Update GaussianRandomFields compat, remove Plots dependency

nefrathenrici:
## Purpose 

This PR removes the implicit dependency on Plots.jl by using the newest GaussianRandomFields release.
```
(EnsembleKalmanProcesses) pkg> st GaussianRandomFields
Project EnsembleKalmanProcesses v2.4.0
Status `~/clima/EnsembleKalmanProcesses.jl/Project.toml`
  [e4b2fa32] GaussianRandomFields v2.2.5

(EnsembleKalmanProcesses) pkg> why Plots
  GaussianRandomFields → Plots
```
Strictly speaking, we don’t need to restrict compatibility, I could manually ensure that I am using GaussianRandomFields 2.2.6 in my workflow. However, I’d prefer to enforce this constraint to prevent Plots from accidentally becoming a dependency, which helps reduce compilation time.

----

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "EnsembleKalmanProcesses"
 uuid = "aa8a2aa5-91d8-4396-bcef-d4f2ec43552d"
 authors = ["CLIMA contributors <clima-software@caltech.edu>"]
-version = "2.4.0"
+version = "2.4.1"
 
 [deps]
 Convex = "f65535da-76fb-5f13-bab9-19810c17039a"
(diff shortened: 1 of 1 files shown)


odunbar (approved):
Besides the tests, I have rerun the example in `examples/Darcy` examples to ensure that this still works fine (as it uses the GaussianRandomField priors).

PS -  we typically like to have a separate PR for the project tags usually, but this should be fine here

PR #489, opened 2025-04-28
New getters for observations etc.

### LINKED ISSUE #487: Improve getters for learning rate and observation
## Issue
- We have a getter `get_\Delta t` , but typically the use case is more for an "algorithm time"
- We have get_obs, get_obs_noise_cov to get the latest observations, but with minibatching etc. we may often require getting observations from past iterations.

## Solution 
- create a getter like `get_cumulative_\Delta t` or `get_time` to get the algorithm time
- add an index argument to `get_obs` or `get_obs_noise_cov` which will select the appropriate minibatched observation.

odunbar:
## Purpose 

Closes #487 

## Content

- added optional `iteration` in getters for observation (e.g., `get_obs(ekp, iteration)` ) also for `get_obs_noise_cov` and `get_obs_noise_cov_inv`
- added some utilities such as `get_minibatch_index(observation_series, iteration)` to also get observations from `iteration`
- added `get_algorithm_time` with accumulates the `get_[\Delta]t(ekp)`
- Unit tests and API docs

----

### DIFF
--- a/docs/src/API/EnsembleKalmanProcess.md
+++ b/docs/src/API/EnsembleKalmanProcess.md
@@ -4,14 +4,24 @@
 CurrentModule = EnsembleKalmanProcesses
 ```
 
+## Primary objects and functions
 ```@docs
 EnsembleKalmanProcess
-FailureHandler
-SampleSuccGauss
-IgnoreFailures
+construct_initial_ensemble
+update_ensemble!
+```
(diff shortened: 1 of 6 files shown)


odunbar:
Add's simple and well-tested functionality, that is backward compatible, merging on passing tests and docs

PR #498, opened 2025-05-22
Add Metadata to `Observation` object

odunbar:
## Purpose 

- Request from @nefrathenrici to extend the utility of `metadata` from `ObservationSeries` to individual `Observation` objects

## Content

- Add optional metadata object to `Observation`, set either with `metadata=...` or `Dict(...,"metadata" => ...)` when creating the `Observation`. It is value `nothing` when not set.
- metadata is stacked under `combine_observations([obs1,obs2])` and is retrieved with `get_metadata(observation)`
- API and additional note in docs provided
- unit tested

----

### DIFF
--- a/src/Observations.jl
+++ b/src/Observations.jl
@@ -284,6 +284,9 @@ struct Observation{
     names::AV4
     "A (vector of) indices of the contained observation blocks"
     indices::AV5
+    "Metadata of any type that the user can group with the Observation"
+    metadata::MD
+
 end
 
 """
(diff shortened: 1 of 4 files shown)


src/Observations.jl
```diff
@@ -473,8 +501,10 @@ function combine_observations(obs_vec::AV) where {AV <: AbstractVector}
     nnew2 = [convert(T, n) for n in nnew]
     T = promote_type((typeof(i) for i in inew)...)
     inew2 = [convert(T, i) for i in inew]
+    T = promote_type((typeof(i) for i in mnew)...)
```
nefrathenrici, line 504:
```suggestion
try
    T = promote_type((typeof(i) for i in mnew)...)
catch
    throw(ArgumentError("Incompatible metadata types detected: $(unique(typeof.(mnew)))"))
end
```
Since metadata can have heterogeneous types, this could fail easily. Could we add an informative warning like this one?

nefrathenrici (approved):
This PR looks good! I am a bit worried about the type promotion but I think just throwing an informative error is enough for the user.

src/Observations.jl
odunbar: This is a good point, let's just scrap it.

PR #503, opened 2025-06-05
[WIP] Fix vector of ParameterDistribution mean bug #494

### LINKED ISSUE #494: Bug throwing error in Sampler constructor
## Issue
If one provides a distribution to `Sampler`, it will extract a mean and cov
https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/28d0c309f9b48854e60334c9ad11338b2b7b3086/src/EnsembleKalmanSampler.jl#L23-L28
Unfortunately, the mean of `Samples`-type distribution is usually a `K-by-1` Matrix, and this throws an error.

This may happen elsewhere too (e.g. UKI?) 

## Quick-fix
Users can pass in `Sampler(mean(distribution)[:], cov(distribution))` instead

## Actual Solution
Vectorize the mean with [:] ? not "Vector(..)"

GabeBolton:
## Purpose 
Closes #494
I have written a test that picks up the bug in issue, and a general solution to `Vector(mean(prior))` failing on K-by-1 priors.

## To-do
There is a possible extension of this fix into the `mean` definition for `VectorOfParameterized` as well.

## Content
As stated in the issue brief, this appears to have at one point affected UKI, and was addressed [the line](https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/0064442056d6e2a970bd1a308234a1ae8db1e97a/src/UnscentedKalmanInversion.jl#L212-L213)
```
    u0_mean = isa(mean(prior), AbstractVector) ? Vector(mean(prior)) : [mean(prior)] # mean of unconstrained distribution
```
I believe flattening as suggested would be more robust. However, I propose that be handled within the [`ParameterDistribution` extension of the `mean` definition](https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/0064442056d6e2a970bd1a308234a1ae8db1e97a/src/ParameterDistributions.jl#L779). This would allow expected behaviour while requiring the fewest changes, and preventing this issue being reintroduced in future development.

Note: there seems to be no dev or `upstream` branch to merge into, if there is a better branch to merge into, please let me know.

----
- [x ] I have read and checked the items on the review checklist.

### DIFF
--- a/src/ParameterDistributions.jl
+++ b/src/ParameterDistributions.jl
@@ -776,7 +776,7 @@ Returns a concatenated mean of the parameter distributions.
 mean(d::Parameterized) = mean(d.distribution)
 mean(d::Samples) = mean(d.distribution_samples, dims = 2)
 mean(d::VectorOfParameterized) = reduce(vcat, mean.(d.distribution))
-mean(pd::ParameterDistribution) = reduce(vcat, mean.(pd.distribution))
+mean(pd::ParameterDistribution) = reduce(vcat, [m[:] for m in mean.(pd.distribution)])
 
 #apply transforms
 function transform_constrained_to_unconstrained(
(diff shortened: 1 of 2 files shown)


odunbar:
#494 now addressed in PR #519

PR #520, opened 2025-09-30
bugfix docs image link

odunbar:
## Purpose 

- New EKS figures load in local docs build but not global: (due to _online-only_ figures linking to `dev/ensemble_kalman_sampler/assets/` and not `dev/assets/` 

## Content

- correct linking
- add new note about passing in "prior, impose_prior=true" to UKI as the recommendation

----


odunbar:
Merging on correct docs preview `https://clima.github.io/EnsembleKalmanProcesses.jl/previews/PR520/` and passing tests

PR #521, opened 2025-10-01
Add fallback for linear solve

nefrathenrici:
Content 
- `safe_linear_solve(A,b)`: solves via `A\b` and falls back to `pinv(A)*b` on `SingularException`, with a warning that can be disabled
- `add_diagonal_regularization!(cov_matrix)`: Adds a `sqrt(eps(eltype(cov_matrix)))` to the input matrix's diagonal 
- Applies `add_diagonal_regularization!` and `safe_linear_solve` to the source code. I am not sure that I applied regularization everywhere it is needed.

### DIFF
--- a/src/LinearSolve.jl
+++ b/src/LinearSolve.jl
@@ -1,0 +1,51 @@
+using LinearAlgebra
+
+export safe_linear_solve, safe_linear_solve!
+
+"""
+    safe_linear_solve(A, b; warn_on_singular=true)
+
+Solves the linear system Ax = b with robust handling of ill-conditioned matrices.
+
+# Arguments
+- `A`: Coefficient matrix
+- `b`: Right-hand side vector or matrix  
+- `warn_on_singular`: Whether to issue warnings when ill-conditioned matrices are detected
+
+# Algorithm
+1. **Primary**: Standard solve `A \\ b` (when matrix is well-conditioned)
+2. **Fallback**: Pseudoinverse solve using normal equations or Moore-Penrose pseudoinverse
+"""
+function safe_linear_solve(
+    A::AbstractMatrix,
+    b::AbstractVecOrMat;
+    warn_on_singular = true,
+    max_cond = 1e12,
+)
+    cond_A = cond(A)
+    is_ill_conditioned = cond_A > max_cond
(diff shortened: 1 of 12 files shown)


src/UnscentedKalmanInversion.jl
```diff
@@ -743,7 +748,10 @@ function update_ensemble_analysis!(
 
     if process.impose_prior
         ug_cov_reg = [ug_cov uu_p_cov]
-        gg_cov_reg = [gg_cov ug_cov'; ug_cov uu_p_cov+process.prior_cov[u_idx, u_idx] / get_Δt(uki)[end]]
+        gg_cov_reg = [
+            gg_cov ug_cov'
+            ug_cov uu_p_cov+process.prior_cov[u_idx, u_idx] / get_Δt(uki)[end]
+        ]
         tmp = ug_cov_reg / gg_cov_reg
```
odunbar, line 755: Should we not add the linear solve here too for the unscented methods?

i think `A / B` is like `(B' \ A')'`

src/LinearSolve.jl
```diff
@@ -0,0 +1,59 @@
...
+
+Solves the linear system Ax = b with robust handling of ill-conditioned matrices.
+
+# Arguments
+- `A`: Coefficient matrix
+- `b`: Right-hand side vector or matrix  
+- `warn_on_singular`: Whether to issue warnings when ill-conditioned matrices are detected
+
+"""
+function safe_linear_solve(A::AbstractMatrix, b::AbstractVecOrMat; warn_on_singular = true)
```
odunbar, line 16: Although this can be used, it appears that `warn_on_singular` is not made settable.
I'd say maybe for now we can link it to "get_verbose(ekp)"?

odunbar (commented):
Hi Nat, 

This looks great thanks! My only comments involve the keyword, and that I believe we have missed applying this in UKI - where the `/` operator was used instead

If you could also add a unit test for this function too I think codecov will be happy.
```
safe_linear_solve!(x::AbstractVecOrMat, A::AbstractMatrix, b::AbstractVecOrMat; kwargs...)
```

odunbar (approved):
LGTM!

PR #535, opened 2025-12-12
Some small fixes to docstrings/docs and small bugs 

### LINKED ISSUE #528: Docs: Add the terminate output in the examples
# Issue

The learning rate scheduler throws a termination condition - mentioned in the Learning Rate Scheduler docs, and some of the example codes, but should be added into the examples on the documentation

### LINKED ISSUE #529: We incorrectly treat +I as in the reduced space when
##Issue

https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/026880cddf0d3b87b991ed8b74345a31e9a29756/src/Observations.jl#L119-L121

Currently we interpret `SVDplusD` with uniform scaling input, as having this act on the SVD reduced space, but this is incorrect. Or rather, it is misleading as users with believe that they are implementing `U*S*V'+I` and not `U*(S+I)*V'`. (which are different)

## Possible solution
We should just treat this as if the uniform scaling as if it was a diagonal matrix.

### LINKED ISSUE #533: Modify GNKI to work with 1D (as done in UKI/EKS)
# Issue

The GNKI fails when creating scalar versions

https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/d5892a4fac9c4a5fb2fb5435696a3762230b6000/src/GaussNewtonKalmanInversion.jl#L17-L21

## Possible solution

This has been resolved for EKS e.g. here: 

https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/d5892a4fac9c4a5fb2fb5435696a3762230b6000/src/EnsembleKalmanSampler.jl#L37-L38

### LINKED ISSUE #534: Docstring incorrect for RMSE
## Issue

The loss metrics computing  `compute_average_unweighted_rmse` and  `compute_average_rmse` compute the RMSE but the docstring shows latex code without the sqrt (i.e. the MSE). this should be fixed as it misleads users

odunbar:
## Purpose 

Closes #534 
Closes #533 
Closes #529 
Closes #528

Local docs build:

## Content

- Fixes GNKI to handle scalar input
- Correctly handles SVDplusD for `aI` case by converting to a Diagonal matrix type
- Fixes the docstrings for RMSE to include the square-roots as this is what is done
- Fixes the docs for a couple of examples, that should use the "terminate" flag, to give users an idea of how it works
- Increases the ensemble size in the sinusoid example so it does not throw warnings.

----

### DIFF
--- a/examples/Sinusoid/sinusoid_example.jl
+++ b/examples/Sinusoid/sinusoid_example.jl
@@ -105,34 +105,10 @@ final_ensemble = get_ϕ_final(prior, ensemble_kalman_process)
 # parameters, the initial ensemble, and the final ensemble.
 p = plot(trange, model(theta_true...), c = :black, label = "Truth", legend = :bottomright, linewidth = 2)
 
-plot!(
-    p,
-    trange,
-    [model(get_ϕ(prior, ensemble_kalman_process, 1)[:, 1]...)],
-    c = :red,
-    label = "Initial ensemble",
-)
-plot!(
-    p,
-    trange,
-    [model(final_ensemble[:, 1]...)],
-    c = :blue,
-    label = "Final ensemble",
-)
-plot!(
-    p,
-    trange,
-    [model(get_ϕ(prior, ensemble_kalman_process, 1)[:, i]...) for i in 2:N_ensemble],
-    c = :red,
-    label = "",
-)
-plot!(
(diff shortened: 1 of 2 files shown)


rgjini (commented):
Everything looks good to me!

Issue #561, opened 2026-05-11
AI-maintainer skills wishlist


odunbar:
## Issue 
I am currently developing some Claude skills for this repository.  I will compile a quick wishlist of some stuff that I think would be great for AI to do, and are currently always on my to-do list. I have a `settings.json` that asks users to use these skills (or create new ones if it is a common task not yet covered)

## Skill philosophy here - enforced by user, and `Meta-skills` below
1. Suited to doing one particular task. We can call on these skills occasionally (and manually for now) like any other tool  to improve the user experience of the repository.
2. Suited for easy copy to different repositories (though tailored to a repository through examples etc.)
3. Must have a mechanism for self-improvement.

## The wishlist 
As I am just learning here, some of these might be ideally suited and others need scope refinement. This set can easily evolve however.

### Meta-skills
- [x] `skill-creator` (adapt anthropic's one)
- [ ] (e.g. something to help the AI navigate the repository efficiently and save tokens)

### Repo-skills
- [x] `docstring` : docstring maintenance and consistency with arguments returns etc. in line with julia conventions, and using DosStringExtensions, listing examples if needed.
- [x] `base-show`: Adding prettified base-show and base-summary methods to all structs in src/. Particularly ones that curb long high-dimensional distributions
- [x] `error-message-manager`: a tool to maintain good consistency of  error messages. Errors should be actionable,.(e.g. Argument errors have `Expected X, received Y.` ). Common errors should be unified into one message, with a method `bad_xyz_call(state)` put into the different sites. in-loop errors must have context (e.g. the iteration and state at which the error was thrown).
- [ ] `slurm-pipeline-manager` for users to a take a local EKP template/pipeline. And write an HPC workflow thatwill create job arrays for the ensemble and handl quirks relating to Julia precompiling etc. or race conditions when writing files 
- [ ] `diagnostic-manager` More consistent reporting of diagnostics for `verbose=true/false`
- [ ] `testset-manager` Better organization of testing, instances and testset encapsulation
- [ ] `unit-test-constructors` Builds organized unit tests for all simple constructors
- [ ] `docs-quicklinks-manager` updates the readme and index quicklinks when docs have changed. 
- [ ] `deprecator` - users can provide an old keyword/api and this will create a deprecated method notice.
- [ ] `readme` and `news` - updates the `Readme.md` or `News.md`

... _add to me_

## Agents
there are also special agent types that might be useful

### the philosophy
[...]

PR #562, opened 2026-05-11
Add some initial Claude settings and skills. Apply these across the repo.

odunbar:
## Purpose 

- Infrastructure and some initial application of the skills for #561 
- I have then applied these and iterated upon them for improved behaviour. Happy in future PRs for greater iteration - in theory almost every application can be used to improve the skill

## Content

- Create Claude `settings.json` - it encourages users to use skills and subagents. It prevents Claude using non-read-type git commands.
- Build some skills
  - `skill-creator` - to build new skills. This was used to build the following from the wishlist:
    - `base-show` adds show and summary methods to structs
    - `docstrings` manages the docstring formatting and manages `docs/src/API`
    - `error-message-manager` manages the error messaging format, to ensure actionable, context-laden messages
- I then used these skills on the repository:
  - `base-show` was used to create show and summary methods for all structs. Includes both a compact and full show-type method (e.g. so that in arrays the compact 1-line show() method is used). unit tests are also written for this.
  - `docstrings` methods was used to overhaul the docstrings.
  -  `error-message-manager` was used to remove `@assert` and update all error messages, it also adds `test_throw` error messages tested in the unit tests. Any longer `>3 line` message is now put into a helper function `_throw_xyz` that are listed at the bottom of files.
- The `skill-creator` ensures that skills have a self-improvement request after every application. Skills were performed a couple of times, using this self-improvement to improve their performance. Also I had it add examples from the repo into skills to help with consistency.
  
## Examples: Current results from `base-show`
Improves vastly upon the spam of 100's of lines that used to happen for high dimensional fields
```
julia> ekiobj
EnsembleKalmanProcess
  process    : Inversion
  N_ens      : 50
  N_par      : 50
  n_iter     : 20
  scheduler  : DataMisfitController
  accelerator: NesterovAccelerator

julia> ekiobj.scheduler
DataMisfitController
  terminate_at: 1.0
  on_terminate: stop
  n_iterations: 1

# demo of the compact show methods .. (from different example)
julia> (ensemble_kalman_process, ensemble_kalman_process)
(EnsembleKalmanProcess (Inversion, N_ens=20, 6 iter), EnsembleKalmanProcess (Inversion, N_ens=20, 6 iter)) 

julia> get_u(ensemble_kalman_process, return_array=false)
7-element Vector{EnsembleKalmanProcesses.DataContainers.DataContainer{Float64}}:
 DataContainer (2×20)
 DataContainer (2×20)
 DataContainer (2×20)
 DataContainer (2×20)
 DataContainer (2×20)
 DataContainer (2×20)
 DataContainer (2×20)
```
```julia
julia> dist = constrained_gaussian("test", 0.3, 5, 0, Inf, repeats = 30)
ParameterDistribution with 1 entry
  'test': VectorOfParameterized (30 distributions) [30 constraints]


julia> dist.distribution
1-element Vector{VectorOfParameterized{Normal{Float64}}}:
 VectorOfParameterized
  n_distributions: 30
  eltype         : Normal{Float64}
```
## Example: Current results from `docstring`
Replaces/adds `DocstringExtensions` where needed
```julia 
"""
An ensemble transform Kalman inversion process.

$(TYPEDEF)

# Fields

$(TYPEDFIELDS)

# Constructors

$(METHODLIST)
"""
```
And 
```julia
"""
$(TYPEDSIGNATURES)

Construct a `Sampler` process from a `ParameterDistribution` prior.

The `sampler_type` keyword selects the update algorithm: `"aldi"` (default) uses the
affine-invariant Langevin dynamics update (`ALDI`); `"eks"` uses the ensemble Kalman
sampler (`EKS`).

# Arguments
- `prior`: parameter prior distribution used to extract the prior mean and covariance.
- `sampler_type`: one of `"aldi"` (default) or `"eks"`.
"""
```
and adds to docs/src/API
```julia
## Update Groups

` ` `@docs
UpdateGroup
get_u_group
get_g_group
get_group_id
update_group_consistency
create_update_groups
` ` `
```

## Example: Current results from `error-message-manager`
It now has a consistent error format, and uses inlined functions to retain readability
```julia
# _throw_xyz() messages
...
if !(eltype(epoch) <: Int)
    _throw_obs_epoch_bad_eltype(eltype(epoch), length(epoch))
end

# ... with context of where it was thrown if ambiguous
...
if !hasmethod(iterate, [typeof(x)]) # no way of iterating
        _throw_x_not_iterable(x; where = :transform_constrained_to_unconstrained)
end

# ... with context if within a loop
for (key, val) in pairs(group_identifiers)
    ...
    for pn in key_vec
        pi = param_indices[pn .== param_names]
        if length(pi) == 0
            _throw_ug_bad_param_name(pn, param_names, key, key_vec)
        end
        ...
    end
   ...    
end
```

with the throw function looking like:
```julia 
@noinline function _throw_obs_epoch_bad_eltype(T, len)
    throw(ArgumentError("""
Epoch vector must contain integer indices.

Expected:
    eltype(epoch) <: Int

Got:
    eltype(epoch) = $T
    length(epoch) = $len

Suggestion:
    Convert the epoch to a Vector{Int}, e.g. `Int.(epoch)`.
"""))
end
```
It also replaces assert statements, with proper errors
```julia
# Replaces: 
@assert(haskey(param_info, "constraint"))

# To:
haskey(param_info, "constraint") || _throw_toml_missing_required_key("constraint", collect(keys(param_info)))

...

@noinline function _throw_toml_missing_required_key(key_name, got_keys)
    throw(ArgumentError("""
Parameter info dict is missing the required $(repr(key_name)) key.

Got keys:
    $got_keys

Suggestion:
    Ensure the TOML entry for this parameter includes a `$key_name = ...` field.
"""))
end
```

----


src/EnsembleKalmanInversion.jl
```diff
@@ -3,9 +3,17 @@
 export get_prior_mean, get_prior_cov, get_impose_prior, get_default_multiplicative_inflation
 
 """
-    Inversion <: Process
+An ensemble Kalman Inversion process.
```
haakon-e, line 6: not sure if this is what you wanted, the name of the struct should be indented on the first line, no?

haakon-e:
The ideas here are really neat and show promise with improving the user experience and code consistency! We should definitely upstream some of these skills that could be useful to other repos too.

Personally, I wouldn't use the `settings.json` rules, as I have other rules/tools and ways to constrain the behavior of the AI, but seems like a very reasonable first-step for many new developers of the repo.

cmbengue (commented):
This looks great--good work. 💯

I especially like providing the AI agents with situation-action and priority-weighted quality criteria.

I've included the following points, which are very minor, for your consideration:
- The error manager reduces code conciseness for increased error clarity, the latter of which is certainly preferable. I wondered whether the same level of clarity could be achieved while reducing code repetition.
- The PR adds the skills, updates the docs, and makes some minor code changes. In the future, separating the skill addition from the implementation might be preferable, as each one has separate but important review questions.
- Would users who access the function documentation by alternative means get the appropriately rendered docstrings?
- Did we lose any important information from the documentation transformation, e.g., the recommended usage in UTKI?

src/EnsembleKalmanInversion.jl
odunbar: In an effort to only use the docstring extensions, I  used the $(TYPEDEF) expression which provides the same information 
e.g.,  from CLI:
```julia
help?> Inversion
search: Inversion pkgversion SparseInversion InsertionSort reverseind TransformInversion

  An ensemble Kalman Inversion process.

  struct Inversion{FT<:AbstractFloat, NorV<:Union{Nothing, AbstractVector}, NorAMorUS<:Union{Nothing, LinearAlgebra.UniformScaling, AbstractMatrix}} <: EnsembleKalmanProcesses.Process

  Fields
  ≡≡≡≡≡≡

    •  prior_mean::Union{Nothing, AbstractVector}: Mean of Gaussian parameter
       prior in unconstrained space

    •  prior_cov::Union{Nothing, LinearAlgebra.UniformScaling, AbstractMatrix}:
       Covariance of Gaussian parameter prior in unconstrained space

    •  impose_prior::Bool: flag to explicitly impose the prior mean and covariance
       during updates

    •  default_multiplicative_inflation::AbstractFloat: if prior is imposed,
       inflation is often required. This sets a default multiplicative inflation
       with s = default_multiplicative_inflation

  Constructors
  ≡≡≡≡≡≡≡≡≡≡≡≡

  Inversion(
      prior_mean,
      prior_cov,
      impose_prior,
      default_multiplicative_inflation
  )
...
...
```

So I could re-order these statements perhaps? Or do you think it is ok?

odunbar:
Thanks for the comments!  

I think I will take the following action:
- For readability of errors: I was considering putting longer error messages into one location (or reduced # locations). Then just having an inline message called `_throw_bad_xyz(context)` inline. I'll see if this works.


@haakon-e 
- a note on the `settings.json` . this file is the only mechanism that can be used to affect the orchestration/wrapper for claude and not just be injected as a prompt to claude. So this is the most effective way to "ban" or "allow" certain usages or add `hooks` external to claude. Claude has been designed build to a [sequenece of `settings.json` files which are merged with a priority order](https://code.claude.com/docs/en/settings). So this setting will override conflicts in `~/.claude/settings.json`, but it can itself be overriden by a local `.claude/settings.local.json`, we can maybe see how this pans out organically? I'd be happy to remove something annoying/redundant. (e.g. the blas thing)

@cmbengue 
- Will definitely reduce in future - in the end creating the skills is partly empirical, so requires multiple iteration for self-improvement. My expectation is perhaps that a clearer separation in future is a `1 skill + its application` type PR. rather than skills-only and application-only PRs. 
- Unsure about rendering. Here the docstrings works for the julia REPL e.g., `?Inversion` or in the docs `API` section. However docstrings require the Julia `DocStringExtension.jl` package, so you would need some julia-aware environment. But this is the case already as we have always used these extensions.
- Although it appears that `UKI/UTKI` settings were lost, they were moved to a more sensible place - namely from the struct definition to the constructor of the struct. Furthermore the `UTKI` details were replaced with the note `accepts the same keywords as Unscented(...) constructor` (which is correct), so in this way we only need to maintain the UKI constructor docstring.

src/EnsembleKalmanInversion.jl
haakon-e: Ah, I see. I'm not opposed to description-first per-se, was just surprised since it's different from e.g. functions. I'd say just leave it for now, it's anyhow trivial to have these AI tools change this in the future if something else is "better".

haakon-e (approved):
Sounds good to try this for now and then update later if need-be.

Thanks for the explanation about the settings.json structure, I was not aware.

.claude/skills/base-show/SKILL.md
```diff
@@ -0,0 +1,389 @@
...
+
+### Step 6 — Offer to improve the skill
+
+After the tests pass and the REPL output looks good, ask the user: "Would you like to improve the **base-show** skill itself using skill-creator? You can suggest changes to the workflow or quality criteria, or I can analyse what came up during this session to identify improvements to the skill."
+
+## Common patterns
+
+### Forward plain show to the MIME method
+
+When a type should look the same whether printed inside a container or directly in the
```
imreddyTeja: This seems like it could result in some strange representations with newlines. Not sure if this would be useful:

" Container types generally implement 3-argument show by calling show(io,
  MIME"text/plain"(), x) for elements x, with :compact => true"

imreddyTeja (approved):
If this ends up working well, we should do something similar in other clima repos

.claude/skills/base-show/SKILL.md
odunbar: This is a good point, the new pattern is designed to overcome this - see examples above

```julia
# 3-arg MIME method: full REPL display, with compact fallback
function Base.show(io::IO, ::MIME"text/plain", x::T)
    if get(io, :compact, false)
        show(io, x)   # delegate to the 2-arg compact method
    else
        println(io, "T")
        println(io, "  field_name : ", summary_value)
        # ...
    end
end

# 2-arg method: single-line compact representation (no newline)
function Base.show(io::IO, x::T)
    print(io, "T (key_hint)")
end
```

odunbar:
Update from review:
- Updated base-show skill and applied it. Now show methods have a "compact" and "full" version. (see examples above)
- Updated error-message-manager skill and applied it. Now messages > 3 lines in length will be moved to the bottom of file in a helper function. 
- Updated `settings.json` to remove the superfluous "Julia" statement

cmbengue (approved):
I think this looks good. The helper functions are a good addition.

PR #564, opened 2026-05-21
Add tests to ensure SVD error decreases monotonically with rank

nefrathenrici:
This PR adds a small unit test to ensure that a truncated SVD's error decreases monotonically as rank increases. No existing functionality is changed.

### DIFF
--- a/test/Observations/runtests.jl
+++ b/test/Observations/runtests.jl
@@ -290,7 +290,26 @@ end
     tall = Dinv * U * S * cholT.L
     @test norm(get_diag_cov(DmT) - Dinv) < 1e-12
     @test norm(get_tall_cov(DmT) - tall) < 1e-12
+end
 
+@testset "tsvd_cov_from_samples monotonic error vs. rank" begin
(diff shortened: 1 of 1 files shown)


odunbar (approved):
I'm fine having a test for this (as in future we may move to approximate SVD solvers) but i'm a bit confused why we need it currently. 

Right now the method `svd(A)` should guarantee this property. Did you see some issues with monotonicity here in your work?

nefrathenrici:
> I'm fine having a test for this (as in future we may move to approximate SVD solvers) but i'm a bit confused why we need it currently.
> 
> Right now the method `svd(A)` should guarantee this property. Did you see some issues with monotonicity here in your work?

I haven't seen any issues with monotonicity, Tapio wanted to ensure that the truncated SVD error decreases monotonically. If this is already guaranteed by an upstream method we don't need to add redundant tests.

odunbar:
It's ok, merge away! Right now it's guaranteed but maybe in future it won't be. I'm ok with redundancy for tests

PR #568, opened 2026-06-01
Throw error for < 2 successful ensemble members

### LINKED ISSUE #565: `update_ensemble!` hangs on all-NaN G_ensemble
When all columns of g are NaN, update_ensemble! hangs indefinitely. Ideally this can throw an error right away. I believe this results in DataMisfitController.calculate_timestep! producing NaN timesteps, leading to sqrt of an all-NaN covariance matrix which never returns. 

minimal reproducer: 
```
using EnsembleKalmanProcesses as EKP, LinearAlgebra

obs_vec = [EKP.Observation(Dict(
    "samples" => randn(60), "covariances" => Diagonal(0.001*ones(60)), "names" => "obs_$i",
)) for i in 1:6]
obs_series = EKP.ObservationSeries(obs_vec, EKP.RandomFixedSizeMinibatcher(6, "trim"), ["obs_$i" for i in 1:6])
ekp = EKP.EnsembleKalmanProcess(randn(10, 100), obs_series, EKP.Inversion();
    scheduler=EKP.DataMisfitController(on_terminate="continue"),
    failure_handler_method=EKP.SampleSuccGauss())
EKP.update_ensemble!(ekp, fill(NaN, 360, 100))  # hangs
```

odunbar:
## Purpose 

- Closes #565 

## Content

- add error throw to `split_indices_by_success` for < 2 ensemble members
- add a test for all-NaN `g` to ensure it throws and not hangs
- add a note in the docs.

----

### DIFF
--- a/docs/src/failure_handling.md
+++ b/docs/src/failure_handling.md
@@ -48,8 +48,8 @@ ukiobj = EnsembleKalmanProcess(
 
 The user determines if a ensemble member has failed and replace the output (column) of ``\mathcal{G}(\theta)`` with `NaN`s. The `FailureHandler` takes care of the rest.
 
-!!! info "Minimum number of successful particles = 3"
-    `update_ensemble!` requires a minimum of 3 successful particles to perform an update. Below this will always throw an error. 
+!!! info "Minimum number of successful particles = 2"
+    `update_ensemble!` requires a minimum of 2 successful particles to perform an update. Below this will always throw an error. 
     
 ### Partial particle failure (a.k.a Imputation)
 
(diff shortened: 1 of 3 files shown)


costachris (approved):
LGTM
3 is a reasonable and extreme bound. Just for my own understanding, is there something special about 3 (vs e.g. > 2 or > 1)?

odunbar:
You know what, we can actually handle 2, I will lower to this for correctness. In my mind we need a min of 3, but we don't!

1 does break covariance computations though obviously. EKI (involves an /(N-1)) for example where N=1

costachris:
sounds good. It's probably also better with CI/tests in mind, where users might just test with a few. There is e.g. a weekly coupler CI calibration with 3 ensemble members.

